PR #18, opened 2022-09-05
Instead of adding our own label, reuse Statefulset selector

krajorama:
This removes the need for the mostly redundant "name" label.

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>


pstibrany:
This change will use all label matchers from statefulset selector to find pods. This is possibly a breaking change. It's not clear to me why it wasn't done before (perhaps due to a scenario when deploying rollout-operator to existing cluster?).

christopherzli:
I am wondering why this is not merged into master.

christopherzli (approved):
LGTM

pracucci:
> This change will use all label matchers from statefulset selector to find pods. This is possibly a breaking change. It's not clear to me why it wasn't done before (perhaps due to a scenario when deploying rollout-operator to existing cluster?).

We need to better investigate this before merging this PR.

christopherzli:
is there any help or anything I can help to confirm in order to merge this PR? In my understanding this would not be a breaking change.

pracucci:
> is there any help or anything I can help to confirm in order to merge this PR? In my understanding this would not be a breaking change.

We need to test it at Grafana Labs to make sure it's not a breaking change for our use cases.

aallawala:
@pracucci, how's testing going at Grafana Labs re: this change?

pracucci:
Priorities changed and we never tested this change.

pracucci:
Closing since I think none will work on this in the foreseeable future. We can always re-open it if things change.

PR #20, opened 2022-09-30
Detect as not-ready pods which are terminating but their readiness probe hasn't failed yet

pracucci:
The rollout-operator checks the non-ready pods looking at the StatefulSet status. Non-ready pods are the ones failing the readiness probe. Mimir ingesters pod readiness probe is configured with `period=10s #success=1 #failure=3`. Given the ingester takes a while to shutdown, the readiness probe doesn't fail immediately but after about 30s (10s x 3 times). Until then, the terminating pod is not reported as non-ready in the StatefulSet status, and so the StatefulSet is not reported has having not ready pods.

This PR improved the logic used to check if a StatefulSet has not ready pods, checking all actual pods status too.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -417,12 +417,12 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
 
 	// Ensure all pods in this StatefulSet are Ready, otherwise we consider a rollout is in progress
 	// (in any case, it's not safe to proceed with other StatefulSets).
-	if sts.Status.Replicas != sts.Status.ReadyReplicas {
+	if hasNotReadyPods, err := c.hasStatefulSetNotReadyPods(sts); err != nil {
+		return true, errors.Wrapf(err, "unable to check if StatefulSet %s has not ready pods", sts.Name)
+	} else if hasNotReadyPods {
 		level.Info(c.logger).Log(
 			"msg", "StatefulSet pods are all updated but StatefulSet has some not-Ready replicas",
-			"statefulset", sts.Name,
-			"replicas", sts.Status.Replicas,
-			"ready_replicas", sts.Status.ReadyReplicas)
+			"statefulset", sts.Name)
 
 		return true, nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -354,12 +417,12 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
 
 	// Ensure all pods in this StatefulSet are Ready, otherwise we consider a rollout is in progress
 	// (in any case, it's not safe to proceed with other StatefulSets).
-	if sts.Status.Replicas != sts.Status.ReadyReplicas {
+	if hasNotReadyPods, err := c.hasStatefulSetNotReadyPods(sts); err != nil {
+		return true, errors.Wrapf(err, "unable to check if StatefulSet %s has not ready pods", sts.Name)
+	} else if hasNotReadyPods {
 		level.Info(c.logger).Log(
```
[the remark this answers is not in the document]
pracucci: It's not fully redundant. The `hasStatefulSetNotReadyPods()` logs only if it founds a discrepancy between the number of not-ready pods reported by the StatefulSet and the actual one. On the contrary, here we always log it.

pkg/controller/util.go
```diff
@@ -25,6 +25,39 @@ func sortPods(pods []*corev1.Pod) {
...
+	}
+
+	// It must not be in the terminating state.
+	if pod.DeletionTimestamp != nil {
+		return false
+	}
+
+	// All containers must be running and ready.
+	for i := len(pod.Status.ContainerStatuses) - 1; i >= 0; i-- {
+		container := pod.Status.ContainerStatuses[i]
```
colega, line 51: Is there any reason why we're traversing this reversed?

colega (approved):

krajorama (approved):
LGTM, it should reduce the race condition

aknuds1 (approved):
Looks good AFAICT.

pkg/controller/util.go
pracucci: No reason. I changed it. It was initially copied from `kubectl` (but logic there is more complicated cause it needs to find 1 "status reason" to show per pod).

PR #39, opened 2023-03-03
Update Go to 1.20.1 and update many dependencies

andyasp:
Updates Go to 1.20.1. I am doing this in preparation of introducing a build image that uses this version.


andyasp:
Currently failing with:
```
panic: Something in this program imports go4.org/unsafe/assume-no-moving-gc to declare that it assumes a non-moving garbage collector, but your version of go4.org/unsafe/assume-no-moving-gc hasn't been updated to assert that it's safe against the go1.20 runtime. If you want to risk it, run with environment variable ASSUME_NO_MOVING_GC_UNSAFE_RISK_IT_WITH=go1.20 set. Notably, if go1.20 adds a moving garbage collector, this program is unsafe to use.
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## main / unreleased
 
+* [ENHANCEMENT] Update Go to `1.20.1`. #38
```
andyasp, line 5:
```suggestion
* [ENHANCEMENT] Update Go to `1.20.1`. #39
```

pracucci (approved):
It's a massive vendor update. I'm not sure I understand if we're doing it for a specific reason or not, but I also don't have a good reason to block, so let's 🤞

andyasp:
> It's a massive vendor update.

Yeah, I did this late on Friday and was surprised at how many dependencies this updated. The k3d `v5.4.6` -> `v5.4.7` caused a cascade. The intent was to stop the panic, but they had updated the `go4.org/unsafe/assume-no-moving-gc` after that release (probably due to timing) so I had to follow it with bumping it directly anyway.

k3d `v5.4.8` was coincidentally released yesterday, so we do have a few options here:

1. Keep the changes as they are now.
2. Go to k3d `v5.4.8` and stomach all the updates if we're doing a big bang anyway.
3. Revert the k3d `v5.4.7` update and try to only bump the indirect dependency.

The unfortunate part of all this is that the `k3d` dependency is only used for the integration tests. I'm slightly inclined to see what option 2 looks like, but I'm open to following another path. I'll add a changelog entry for whatever direct dependencies end up getting updated.

andyasp:
Attempting the k3d `v5.4.8` update didn't go smoothly so I reverted it.

andyasp:
Merging as-is and we can test it in a dev cell afterwards.

PR #41, opened 2023-03-06
More generous timeouts for expiring cert e2e tests

colega:
CI is slow, and it takes a while to restart the pod, update the secret,
etc. Let's make it generous to make sure it doesn't fail there.

Signed-off-by: Oleg Zaytsev <mail@olegzaytsev.com>

### DIFF
--- a/integration/e2e_test.go
+++ b/integration/e2e_test.go
@@ -289,7 +289,7 @@ func TestExpiringCertificate(t *testing.T) {
 		require.Eventuallyf(t, func() bool {
 			currentSecretCertExpiration := getCertificateExpirationFromSecret(t, api, ctx)
 			return currentSecretCertExpiration.After(secretCertExpiration)
-		}, 30*time.Second, 1*time.Second, "certificate should be renewed and expiration date should change")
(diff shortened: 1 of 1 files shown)


colega:
CI failed with 
```
panic: Something in this program imports go4.org/unsafe/assume-no-moving-gc to declare that it assumes a non-moving garbage collector, but your version of go4.org/unsafe/assume-no-moving-gc hasn't been updated to assert that it's safe against the go1.20 runtime. If you want to risk it, run with environment variable ASSUME_NO_MOVING_GC_UNSAFE_RISK_IT_WITH=go1.20 set. Notably, if go1.20 adds a moving garbage collector, this program is unsafe to use.
```

:facepalm: 

_Packages that explicitly break golang's compatibility promise_

Need to update k3d to fix this, and apparently that will require fixing more deps.

colega:
I will (try to) fix the dependencies in a separate PR first.

andyasp:
@colega I hit the same issue and did a k3d update in #39 so you may just need to undo the update here and rebase.

colega:
@andyasp thank you for taking care of that.

Rebased and force-pushed this, dropping the vendoring commit.

colega:
CI is still flaky, need to take another look.

PR #46, opened 2023-03-28
Makefile: Set dependencies for binary

aknuds1:
In Makefile, set the dependencies for the rollout-operator binary so it gets re-built when Go files change.

I've copied and adapted the logic from the Mimir Makefile for finding Go source files. Please review whether this logic makes sence, since it's a bit hairy :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,10 @@ IMAGE_TAG ?= $(GIT_BRANCH)-$(GIT_REVISION)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 
-rollout-operator:
+DONT_FIND := -name vendor -prune -o -name .git -prune -o -name .cache -prune -o -name .pkg -prune -o
+GO_FILES := $(shell find . $(DONT_FIND) -name cmd -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)
+
+rollout-operator: $(GO_FILES)
 	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=0 go build -ldflags '-extldflags "-static"' ./cmd/rollout-operator
 
 .PHONY: build-image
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -7,7 +7,10 @@ IMAGE_TAG ?= $(GIT_BRANCH)-$(GIT_REVISION)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 
-rollout-operator:
+DONT_FIND := -name vendor -prune -o -name .git -prune -o -name .cache -prune -o -name .pkg -prune -o
+GO_FILES := $(shell find . $(DONT_FIND) -name cmd -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)
```
andyasp, line 11:
```suggestion
GO_FILES := $(shell find . $(DONT_FIND) -type f -name '*.go' -print)
```
I think these prunes can be removed.

As I understand it the intent of `-name cmd -prune -o` is to ignore irrelevant code when there are multiple directories in `cmd` but something else should include the directory that matters, but the latter half of that is missing here so modifying `cmd/rollout-operator/main.go` doesn't trigger a rebuild. There's only `cmd/rollout-operator` here anyway so it seems less confusing to leave it out.

There are currently no `'*.pb.go'` files outside of `vendor`, which is already pruned.

andyasp (approved):
Thanks! Minor suggestion

Makefile
aknuds1: Thank you!

Makefile
```diff
@@ -7,7 +7,10 @@ IMAGE_TAG ?= $(GIT_BRANCH)-$(GIT_REVISION)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 
-rollout-operator:
+DONT_FIND := -name vendor -prune -o -name .git -prune -o -name .cache -prune -o -name .pkg -prune -o
+GO_FILES := $(shell find . $(DONT_FIND) -type f -name '*.go' -print)
```
krajorama, line 11: Nit:

```suggestion
DONT_FIND := -name vendor -prune -o -name .git -prune -o -name .cache -prune -o -name .pkg -prune
GO_FILES := $(shell find . $(DONT_FIND) -o -type f -name '*.go' -print)
```

PR #47, opened 2023-03-31
Prepare downscale webhook

JordanRushing:
This PR follows https://github.com/grafana/rollout-operator/pull/25 and implements a `prep_downscale` mutating admission webhook that, after receiving a request that scales down a resource, sends a HTTP POST to pods in a StatefulSet/Deployment/ReplicaSet that are labeled with `grafana.com/prep-downscale: true`.

The endpoint targets for the requests are crafted by extracting and combining the `grafana.com/prep-downscale-http-path` and `grafana.com/prep-downscale-http-port` labels along with the diff in the number of replicas.

An example scenario for when this is useful:
- KEDA evaluates a metric query and suggests to scale down the replicas in a StatefulSet
- The pods in the StatefulSet implement a `prep_shutdown` HTTP endpoint that tells them to expect to be terminated on the next SIGTERM and manage their state appropriately
- The new mutating admission webhook activates, validates the scale down, successfully hits the `prep_shutdown` HTTP endpoints on relevant pods, and allows the admission request thus resulting in the pods being deleted gracefully


colega:
👋  Thank you for taking this task.

Did a quick first review, and have some feedback:
- This has to be a MutatingWebhook, and it should state that it has side-effects.
- We should add an annotation to the downscaled resource saying when it was downscaled last time, so we can validate that and disallow downscaling other zones when one of the zones was recently downscaled.

pkg/admission/prep_downscale_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+type templateParams struct {
+	Replicas         int
+	DownScalePathKey string
+	DownScalePath    string
+	DownScalePortKey string
+	DownScalePort    string
+}
+
+func testPrepDownscaleWebhook(t *testing.T, oldReplicas, newReplicas, httpStatusCode int, allowed bool, podsPrepared bool) {
```
colega, line 54: Can we write an integration test please?

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			return allowBecauseCannotGetResource(ar, logger, err)
+		}
+	default:
+		return allowWarn(logger, fmt.Sprintf("unsupported type %T, allowing the change", o))
+	}
+
+	// Since it's a downscale, check if the resource has the label that indicates it's ready to be prepared to be downscaled.
+	// Create a slice of endpoint addresses for pods to send HTTP post requests to and to fail if any don't return 200
+	if lbls[PrepDownscaleLabelKey] == PrepDownscaleLabelValue {
```
colega, line 96: We need to make sure it's not a dry run.

MichelHollands: This has been added.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,185 @@
...
+					level.Error(logger).Log("msg", "error sending HTTP post request", "endpoint", ep, "err", err)
+					return err
+				}
+				if resp.StatusCode != 200 {
+					err := fmt.Errorf("HTTP post request returned non-200 status code")
+					level.Error(logger).Log("msg", err, "endpoint", ep, "status", resp.StatusCode)
+					return err
+				}
+				if resp.StatusCode == 200 {
+					// TODO(jordanrushing): set a label on the pod to indicate it's been prepared for shutdown? Or do we need to do that?
```
56quarters, line 136: I think it's a good idea. There was some questions about how to persist the fact that a pod has been asked to prepare for downscaling. E.g. if a pod prepares for downscaling, then crashes, restarts and doesn't realize it has been asked to prepare for downscaling, is shutdown but _doesn't_ do any of the special "prepare for downscale" logic.

The situation above requires the application to know it's running in Kubernetes and inspect annotations which isn't great but I'm not sure there's a good way around that.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,185 @@
...
+			ep := ep // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				resp, err := client.Post("http://"+ep, "application/json", nil)
+				if err != nil {
+					level.Error(logger).Log("msg", "error sending HTTP post request", "endpoint", ep, "err", err)
+					return err
+				}
+				if resp.StatusCode != 200 {
+					err := fmt.Errorf("HTTP post request returned non-200 status code")
+					level.Error(logger).Log("msg", err, "endpoint", ep, "status", resp.StatusCode)
```
56quarters, line 132: Should probably include the body in the log message, right? Seems like this would be kinda hard to debug otherwise.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,185 @@
...
+					return err
+				}
+				if resp.StatusCode == 200 {
+					// TODO(jordanrushing): set a label on the pod to indicate it's been prepared for shutdown? Or do we need to do that?
+					// If we did that, when the rollout operator reconcile loop runs, it would need to "de-prep" the pod and remove the label
```
56quarters, line 137: I was under the impression that pods that have been prepared for shutdown are going to be downscaled and hence deleted (so there's no cleanup required).

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,185 @@
...
+	case *autoscalingv1.Scale:
+		lbls, err = getResourceLabels(ctx, ar, api)
+		if err != nil {
+			return allowBecauseCannotGetResource(ar, logger, err)
+		}
+	default:
+		return allowWarn(logger, fmt.Sprintf("unsupported type %T, allowing the change", o))
+	}
+
+	// Since it's a downscale, check if the resource has the label that indicates it's ready to be prepared to be downscaled.
```
56quarters, line 100:
```suggestion
	// Since it's a downscale, check if the resource has the label that indicates it needs to be prepared to be downscaled.
```

56quarters: If I understand correct, I believe this is more accurate (I might be entirely wrong)

pkg/admission/prep_downscale.go
MichelHollands: We can use the https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ to expose it as an environment variable.

colega: I think it's interesting to set an _annotation_, not really a label.

>  E.g. if a pod prepares for downscaling, then crashes, restarts and doesn't realize it has been asked to prepare for downscaling, is shutdown but doesn't do any of the special "prepare for downscale" logic.

I described this scenario in the (internal) design doc, and yes, the ingester should somehow persist the fact that it's prepared for downscale.

> We can use the https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ to expose it as an environment variable.

Unless there's something I'm missing, this would require an ingester restart, which we don't want to do here.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+	// Create a slice of endpoint addresses for pods to send HTTP post requests to and to fail if any don't return 200
+	if !*ar.Request.DryRun && lbls[PrepDownscaleLabelKey] == PrepDownscaleLabelValue {
```
colega, line 102: Nit, but I'd do a shortcut here:

```go
	if lbls[PrepDownscaleLabelKey] != PrepDownscaleLabelValue {
		// Not labeled, nothing to do.
		return &v1.AdmissionResponse{Allowed: true}
	}
	if *ar.Request.DryRun {
		// TODO: check whether downscale is possible (was another zone downscaled less than allowed time before?)
		// Respond based on that.
	}	
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+		// The service in this case is ingester-zone-a as well.
+		// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
+		for i := 0; i < int(diff); i++ {
+			eps[i].url = fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local:%s/ingester/%s",
+				ar.Request.Name,       // pod name
+				int(*oldReplicas)-i-1, // nr in statefulset
+				ar.Request.Name,       // svc name
+				ar.Request.Namespace,
+				lbls[PrepDownscalePortKey],
+				lbls[PrepDownscalePathKey],
```
colega, line 121: We should check first whether these labels are set and have valid values, and fail it not.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+		eps := make([]endpoint, diff)
+
+		// The DNS entry for a pod of a stateful set is
+		// ingester-zone-a-0.$(servicename).$(namespace).svc.cluster.local
+		// The service in this case is ingester-zone-a as well.
+		// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
+		for i := 0; i < int(diff); i++ {
+			eps[i].url = fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local:%s/ingester/%s",
+				ar.Request.Name,       // pod name
+				int(*oldReplicas)-i-1, // nr in statefulset
```
colega, line 117: Nit: I'd extract this into a var, since it's used twice.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		g, _ := errgroup.WithContext(ctx)
+		for _, ep := range eps {
+			ep := ep // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				resp, err := client.Post("http://"+ep.url, "application/json", nil)
+				if err != nil {
+					level.Error(logger).Log("msg", "error sending HTTP post request", "endpoint", ep.url, "err", err)
+					return err
```
colega, line 133: Please wrap these errors with ep.index and/or ep.url to simplify debugging.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+				if resp.StatusCode != 200 {
+					err := fmt.Errorf("HTTP post request returned non-200 status code")
+					level.Error(logger).Log("msg", err, "endpoint", ep.url, "status", resp.StatusCode)
+					return err
+				}
+				if resp.StatusCode == 200 {
```
colega, line 140: This condition is inferred from the one above.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+					// TODO(jordanrushing): set a label on the pod to indicate it's been prepared for shutdown? Or do we need to do that?
+					// If we did that, when the rollout operator reconcile loop runs, it would need to "de-prep" the pod and remove the label
+					// OR
+					// we could do that in the conditional below
+					level.Debug(logger).Log("msg", "pod prepared for shutdown", "endpoint", ep.url)
+
+					err = addDownscaledAnnotationToPod(ctx, api, ar.Request.Namespace, ar.Request.Name, ep.index)
```
colega, line 147: Should this be `addPreparedToBeDownscaled`? We didn't downscale yet.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,178 @@
...
+					level.Debug(logger).Log("msg", "annotation added to pod", "err", err)
```
colega, line 148: Log which pod, please, and change the message if `err != nil`

```suggestion
					level.Debug(logger).Log("msg", "annotation added to pod", "err", err, "url", ep.url)
```

pkg/admission/prep_downscale.go
56quarters: > Unless there's something I'm missing, this would require an ingester restart, which we don't want to do here.

Looks like the downward API can also be used to [populate files](https://kubernetes.io/docs/concepts/storage/volumes/#downwardapi) and supports pod annotations. I'll look into this further and see if we can use it.

56quarters: Confirmed that this will work in our environment. Changes to pod annotations are reflected in the on-disk file.

```
~ # cat /etc/podinfo/annotations 
cni.projectcalico.org/containerID="<redacted>"
cni.projectcalico.org/podIP="<redacted>/32"
cni.projectcalico.org/podIPs="<redacted>/32"
kubernetes.io/config.seen="2023-04-10T16:20:28.230821451Z"
kubernetes.io/config.source="api"
```

pkg/admission/prep_downscale.go
MichelHollands: Added

pkg/admission/prep_downscale.go
MichelHollands: The comment has been removed.

pkg/admission/prep_downscale.go
MichelHollands: Applied manually.

pkg/admission/prep_downscale.go
MichelHollands: Good idea. This has been implemented as such.

pkg/admission/prep_downscale.go
MichelHollands: Checks have been added.

pkg/admission/prep_downscale.go
MichelHollands: Good point. A var was added.

pkg/admission/prep_downscale.go
MichelHollands: Added to the logger.

pkg/admission/prep_downscale.go
MichelHollands: Indeed, the if statement has been removed.

pkg/admission/prep_downscale.go
MichelHollands: Yes, that is cleaner. It has been renamed.

pkg/admission/annotations.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	LastDownscaleAnnotationKey = "last-prepared-for-downscale"
```
56quarters, line 15: Can we prefix and name this similar to how the "no downscale" hook works? `grafana.com/prepared-for-downscale`?

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,208 @@
...
+			)
+			eps[i].index = index
+		}
+
+		g, _ := errgroup.WithContext(ctx)
+		for _, ep := range eps {
+			ep := ep // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				logger = log.With(logger, "url", ep.url, "index", ep.index)
+				resp, err := client.Post("http://"+ep.url, "application/json", nil)
```
56quarters, line 162: Should we be using a timeout or retry logic here?

pkg/admission/annotations.go
```diff
@@ -0,0 +1,44 @@
...
+		return fmt.Errorf("multiple or no pods found for statefulset %v and index %v", stsName, podNr)
+	}
+
+	pod := pods.Items[0]
+
+	annotations := pod.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[LastDownscaleAnnotationKey] = time.Now().UTC().String()
```
56quarters, line 39: For my own clarification, just the presence of this annotation is what matters to a running application, right? The fact that the annotation exists means this pod should expect to be permanently shut down. The date here is for the benefit of us (the operators). If this is the case, could you note that in comments or the README somewhere.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,208 @@
...
+				if resp.StatusCode != 200 {
+					err := errors.New("HTTP post request returned non-200 status code")
+					body, readError := io.ReadAll(resp.Body)
+					defer resp.Body.Close()
+					level.Error(logger).Log("msg", err, "status", resp.StatusCode, "response_body", body)
+					return errors.Join(err, readError)
+				}
+				level.Debug(logger).Log("msg", "pod prepared for shutdown")
+
+				err = addPreparedForDownscaleAnnotationToPod(ctx, api, ar.Request.Namespace, ar.Request.Name, ep.index)
```
56quarters, line 176: I think the annotation should be added to the pod _before_ the endpoint is called. This avoids a potential race condition:
* The "prep shutdown" endpoint is called, the pod / ingester in question modifies its in-memory configuration accordingly.
* The periodic task that uses the [downwardAPI](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) to read annotations from a file on disk runs after the "prep shutdown" endpoint is called but before an annotation has been added to the pod, reverting the configuration change made by the "prep shutdown" endpoint.

Changing the order to add the annotation first and then call the endpoint means that the result of calling the endpoint won't be reverted by the periodic job.

pkg/admission/annotations.go
MichelHollands: Done

pkg/admission/annotations.go
MichelHollands: A comment was added.

pkg/admission/prep_downscale.go
MichelHollands: Good point. This was added.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,208 @@
...
+		for _, ep := range eps {
+			ep := ep // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				logger = log.With(logger, "url", ep.url, "index", ep.index)
+				resp, err := client.Post("http://"+ep.url, "application/json", nil)
+				if err != nil {
+					level.Error(logger).Log("msg", "error sending HTTP post request", "err", err)
+					return err
+				}
+				if resp.StatusCode != 200 {
```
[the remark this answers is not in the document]
MichelHollands: Done.

pkg/admission/prep_downscale_test.go
MichelHollands: As this PR is getting bigger and bigger I propose to do this in a separate PR.

56quarters (approved):
I think this is in a good place to merge to main for further testing. LGTM unless @colega would like to see any other changes.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,211 @@
...
+		return allowWarn(logger, fmt.Sprintf("unsupported type %T, allowing the change", o))
+	}
+
+	if lbls[PrepDownscaleLabelKey] != PrepDownscaleLabelValue {
+		// Not labeled, nothing to do.
+		return &v1.AdmissionResponse{Allowed: true}
+	}
+
+	port, ok := lbls[PrepDownscalePortKey]
+	if !ok || len(port) == 0 {
```
colega, line 110: Nit, but can be just:
```suggestion
	port := lbls[PrepDownscalePortKey]
	if port == "" {
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,211 @@
...
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because the %v label is not set or empty.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, PrepDownscalePortKey),
+			},
+		}
+		level.Warn(logger).Log("msg", fmt.Sprintf("downscale not allowed because the %v label is not set or empty", PrepDownscalePortKey))
+		return &reviewResponse
+	}
+
+	path, ok := lbls[PrepDownscalePathKey]
+	if !ok || len(path) == 0 {
```
colega, line 122: Also can be:

```suggestion
	path := lbls[PrepDownscalePathKey]
	if path == "" {
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,211 @@
...
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because the %v label is not set or empty.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, PrepDownscalePathKey),
+			},
+		}
+		level.Warn(logger).Log("msg", fmt.Sprintf("downscale not allowed because the %v label is not set or empty", PrepDownscalePathKey))
+		return &reviewResponse
+	}
+
+	// Since it's a downscale, check if the resource has the label that indicates it needs to be prepared to be downscaled.
+	// Create a slice of endpoint addresses for pods to send HTTP post requests to and to fail if any don't return 200
+	if !*ar.Request.DryRun {
```
colega, line 135: I'd invert this condition, and do an early return in case of dry run, to keep the indentation at minimum.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,211 @@
...
+		g, _ := errgroup.WithContext(ctx)
+		for _, ep := range eps {
+			ep := ep // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				logger = log.With(logger, "url", ep.url, "index", ep.index)
+				err = addPreparedForDownscaleAnnotationToPod(ctx, api, ar.Request.Namespace, ar.Request.Name, ep.index)
+				if err != nil {
+					level.Debug(logger).Log("msg", "adding annotation to pod failed", "err", err)
+				}
+				level.Debug(logger).Log("msg", "annotation added to pod")
```
colega, line 169: What is the purpose of annotating the pod? We can't even trust this annotation, as it will be removed if the pod is rescheduled.

colega (commented):
This comment still applies:

> We should add an annotation to the downscaled resource saying when it was downscaled last time, so we can validate that and disallow downscaling other zones when one of the zones was recently downscaled.

We're doing this to make possible the automation of downscaling, and we need to have the minimum safety net here that will prevent zones from being downscaled with a too small interval.

Also, left some other comments:
- We need to properly check the dry-run (with the condition above)
- I don't see much point in annotating the pods, because that annotation isn't persistent (it's lost if the pod is moved).

pkg/admission/prep_downscale.go
MichelHollands: These are there for restarts of the pod. If this happens before the pod could be shutdown it will start up again without `prepare_shutdown` being called on it. https://github.com/grafana/loki/pull/9153 will use it's presence to call `prepare_shutdown`.

MichelHollands:
> This comment still applies:
> 
> > We should add an annotation to the downscaled resource saying when it was downscaled last time, so we can validate that and disallow downscaling other zones when one of the zones was recently downscaled.
> 
> We're doing this to make possible the automation of downscaling, and we need to have the minimum safety net here that will prevent zones from being downscaled with a too small interval.

@colega By `downscaled resource` you mean the stateful set?

colega:
@MichelHollands 

> @colega By downscaled resource you mean the stateful set?

Yes, I mean the statefulset. We have 3 zones, and we need to wait a safety period of time between downscaling each one of them (the three statefulsets sharing the zone label).

Maybe this isn't a requirement for Loki and we should implement it optional for Mimir? Anyway, I'd define the waiting period as a label on the statefulset too (`grafana.com/min-time-between-zones-downscale: 12h`)

(OMG the day was long, and I edited your comment instead of replying :facepalm:)

MichelHollands:
> Yes, I mean the statefulset. We have 3 zones, and we need to wait a safety period of time between downscaling each one of them (the three statefulsets sharing the zone label).
> 
> Maybe this isn't a requirement for Loki and we should implement it optional for Mimir? Anyway, I'd define the waiting period as a label on the statefulset too (`grafana.com/min-time-between-zones-downscale: 12h`)
 
Yeah, in Loki we don't need this time between downscaling. The rollout operator will make sure only 1 stateful set at a time is scaled up or down. The Loki ingesters will have had the prep-downscale endpoint called. So this works for us.

colega:
> The rollout operator will make sure only 1 stateful set at a time is scaled up or down. 

The rollout operator does not participate in the scaling down operation.

pkg/admission/annotations.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"time"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	LastDownscaleAnnotationKey = "grafana.com/last-prepared-for-downscale"
```
colega, line 14:
```suggestion
	LastDownscaleAnnotationKey = "grafana.com/last-downscale"
```

MichelHollands: @colega I just added a check for other zones having been rolled out recently. This suggestion was applied manually later.

pkg/admission/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	LastDownscaleAnnotationKey             = "grafana.com/last--downscale" // Should be in time.RFC3339 format
```
colega, line 14:
```suggestion
	LastDownscaleAnnotationKey             = "grafana.com/last-downscale" // Should be in time.RFC3339 format
```

pkg/admission/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+	TimeBetweenZonesDownscaleAnnotationKey = "grafana.com/min-time-between-zones-downscale"
```
colega, line 15:
```suggestion
	TimeBetweenZonesDownscaleLabelKey = "grafana.com/min-time-between-zones-downscale"
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,246 @@
...
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because the %v label is not set or empty.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, PrepDownscalePathKey),
+			},
+		}
+		level.Warn(logger).Log("msg", fmt.Sprintf("downscale not allowed because the %v label is not set or empty", PrepDownscalePathKey))
+		return &reviewResponse
+	}
+
+	if *ar.Request.DryRun {
+		return &v1.AdmissionResponse{Allowed: true}
+	}
```
colega, line 136: This should be after we checked the "min time ago" condition.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,246 @@
...
+
+	rolloutGroup := lbls[RolloutGroupLabelKey]
+	if rolloutGroup == "" {
+		// Not part of a rollout group, nothing to do.
+		return &v1.AdmissionResponse{Allowed: true}
+	}
```
colega, line 142: I think this is not correct. Someone labeled this statefulset as it should be prepared for downscale. It's not part of a rollout group, so there's no need to check other zones, but we still want to prepare it for downscale.
I'd say that:
```
if rolloutGroup != "" {
    // ... check other zones last downscale
}
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,246 @@
...
+			},
+		}
+		level.Warn(logger).Log("msg", "downscale not allowed due to error while adding annotation", "err", err)
+		return &reviewResponse
+	}
+	if found {
+		reviewResponse := v1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because another statefulset was downscaled recently.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas),
```
colega, line 158: I would change `findDownscalesDoneMinTimeAgo` to provide enough information for this logline to say:

```
because statefulset %s was downscaled at %s and is labeled to wait %s between zone downscales.
```

This would provide a meaningful message to the caller.

pkg/admission/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		lastDownscaleLabel, ok := sts.Labels[LastDownscaleAnnotationKey]
+		if !ok {
+			// No last downscale label set on the statefulset, we can continue
+			continue
+		}
+
+		downscaleTime, err := time.Parse(time.RFC3339, lastDownscaleLabel)
+		if err != nil {
+			return false, err
```
colega, line 61:
```suggestion
			return false, fmt.Errorf("can't parse last downscale time of %s: %s", sts.Name, err)
```

pkg/admission/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+	if err != nil {
+		return false, err
+	}
+
+	for _, sts := range list.Items {
+		if sts.Name == stsName {
+			continue
+		}
+		lastDownscaleLabel, ok := sts.Labels[LastDownscaleAnnotationKey]
```
colega, line 53: This is weird, we should be checking annotations, right?
```suggestion
		lastDownscaleAnnotation, ok := sts.Annotations[LastDownscaleAnnotationKey]
```

pkg/admission/annotations.go
```diff
@@ -0,0 +1,81 @@
...
+			// No time between downscale label set on the statefulset, we can continue
+			continue
+		}
+
+		timeBetweenDownscale, err := time.ParseDuration(timeBetweenDownscaleLabel)
+		if err != nil {
+			return false, err
+		}
+
+		if downscaleTime.Add(timeBetweenDownscale).After(time.Now().UTC()) {
```
colega, line 75: Nit: there's no need to call `.UTC()` here.

pkg/admission/annotations.go
MichelHollands: Good idea. I've applied this manually.

pkg/admission/prep_downscale.go
MichelHollands: Yes indeed. It has been moved.

pkg/admission/prep_downscale.go
MichelHollands: Indeed, that is a scenario I hadn't thought off.

pkg/admission/prep_downscale.go
MichelHollands: The function now returns a struct with information about the found statefulset. It uses that to fill in the error message as suggested.

pkg/admission/annotations.go
MichelHollands: Good idea. A similar change was made for the duration parsing a bit further.

pkg/admission/annotations.go
MichelHollands: My mistake, I changed that wrongly. Now `grafana.com/last-downscale` is an annotation as it is set by the webhook. `grafana.com/min-time-between-zones-downscale` is a label as it is set by us when the stateful set is created.

MichelHollands:
@colega I think I've addressed all the review comments. Can you have another look?
The CI failure seems unrelated: the pod is pending. Can you rerun the CI? I don't have permission.

colega (approved):
This is looking very good! Thank you for addressing all my feedback. Left one last comment about updating operation, but otherwise LGTM!

pkg/admission/annotations.go
```diff
@@ -0,0 +1,84 @@
...
+
+		timeBetweenDownscaleLabel, ok := sts.Labels[MinTimeBetweenZonesDownscaleLabelKey]
+		if !ok {
+			// No time between downscale label set on the statefulset, we can continue
+			continue
+		}
+
+		timeBetweenDownscale, err := time.ParseDuration(timeBetweenDownscaleLabel)
+		if err != nil {
+			return nil, fmt.Errorf("can't parse %v label of %s: %s", MinTimeBetweenZonesDownscaleLabelKey, sts.Name, err)
```
DylanGuedes, line 70:
```suggestion
			return nil, fmt.Errorf("can't parse %v label of %s: %w", MinTimeBetweenZonesDownscaleLabelKey, sts.Name, err)
```

pkg/tlscert/webhookconfig.go
```diff
@@ -64,3 +64,54 @@ func PatchCABundleOnValidatingWebhooks(ctx context.Context, logger log.Logger, k
...
+			level.Info(logger).Log("msg", "mutating webhook configuration already has same CA bundle set, not patching", "name", wh.GetName())
+			continue
+		}
+
+		data, err := json.Marshal(wh)
+		if err != nil {
+			return fmt.Errorf("failed to marshal mutating webhook configuration: %w", err)
+		}
+
+		res, err := kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Patch(context.Background(), wh.GetName(), types.MergePatchType, data, metav1.PatchOptions{})
```
DylanGuedes, line 109:
```suggestion
		res, err := kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Patch(ctx, wh.GetName(), types.MergePatchType, data, metav1.PatchOptions{})
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+			resp, err := client.Post("http://"+ep.url, "application/json", nil)
+			if err != nil {
+				level.Error(logger).Log("msg", "error sending HTTP post request", "err", err)
+				return err
+			}
+			if resp.StatusCode/100 != 2 {
+				err := errors.New("HTTP post request returned non-2xx status code")
+				body, readError := io.ReadAll(resp.Body)
+				defer resp.Body.Close()
+				level.Error(logger).Log("msg", err, "status", resp.StatusCode, "response_body", body)
```
DylanGuedes, line 204: how large can be the response_body? depending on the size we might prefer to not log it

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+		eps[i].url = fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local:%s/ingester/%s",
+			ar.Request.Name, // pod name
+			index,
+			ar.Request.Name, // svc name
+			ar.Request.Namespace,
+			port,
+			path,
+		)
+		eps[i].index = index
+	}
```
DylanGuedes, line 187: WDYT?
```suggestion
	for i := int(*newReplicas); i < int(*oldReplicas); i++ {
		eps[i].url = fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local:%s/ingester/%s",
			ar.Request.Name, // pod name
			index,
			ar.Request.Name, // svc name
			ar.Request.Namespace,
			port,
			path,
		)
		eps[i].index = index
	}
```

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+	err = g.Wait()
+	if err != nil {
+		// Down-scale operation is disallowed because a pod failed to prepare for shutdown and cannot be deleted
+		reviewResponse := v1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because one or more pods failed to prepare for shutdown.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas),
+			},
+		}
+		level.Warn(logger).Log("msg", "downscale not allowed due to error", "err", err)
```
DylanGuedes, line 220:
```suggestion
		level.Error(logger).Log("msg", "downscale not allowed due to error", "err", err)
```

pkg/admission/prep_downscale.go
DylanGuedes, line 204: WDYT of having a dedicated "err" field for the err? Like:
("msg", "non-2xx received", "err", err, ...)

motiavtion is: it would allow us of searching for errors precisely by doing `{mystream} |= "err="`

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+
+	err = addDownscaledAnnotationToStatefulSet(ctx, api, ar.Request.Namespace, ar.Request.Name)
+	if err != nil {
+		reviewResponse := v1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because adding an annotation to the statefulset failed.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas),
+			},
+		}
+		level.Warn(logger).Log("msg", "downscale not allowed due to error while adding annotation", "err", err)
```
DylanGuedes, line 232:
```suggestion
		level.Error(logger).Log("msg", "downscale not allowed due to error while adding annotation", "err", err)
```

pkg/admission/annotations.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		lastDownscaleAnnotation, ok := sts.Annotations[LastDownscaleAnnotationKey]
+		if !ok {
+			// No last downscale label set on the statefulset, we can continue
+			continue
+		}
+
+		downscaleTime, err := time.Parse(time.RFC3339, lastDownscaleAnnotation)
+		if err != nil {
+			return nil, fmt.Errorf("can't parse %v annotation of %s: %s", LastDownscaleAnnotationKey, sts.Name, err)
```
DylanGuedes, line 59:
```suggestion
			return nil, fmt.Errorf("can't parse %v annotation of %s: %w", LastDownscaleAnnotationKey, sts.Name, err)
```

DylanGuedes (changes requested):
looking good, just a few questions/suggestions

andyasp (commented):
Looks like this is getting close to being merged. Remember to add an entry to the CHANGELOG (and sorry for the previous dependency update conflict!)

pkg/admission/prep_downscale.go
MichelHollands: For Loki and Mimir at the moment it will be empty. They will log the error in more detail so you have to look there.

pkg/admission/prep_downscale.go
MichelHollands: That doesn't define a field called index. Do you mean using `i` instead of `index`?

pkg/admission/prep_downscale.go
MichelHollands: Yes, that makes it consistent with the other error log messages as well.

pkg/admission/prep_downscale.go
DylanGuedes: yep, mostly because iterating from 0 to diff and evaluating index as `oldreplicas - i - 1`, which is a little confusing, just iterate from newReplicas to oldReplicas.

MichelHollands:
> Looks like this is getting close to being merged. Remember to add an entry to the CHANGELOG (and sorry for the previous dependency update conflict!)

The CHANGELOG has been updated as well as the README.md. The e2e tests were fixed and work consistently locally. However in the CI the last test fails occasionally. I'm not sure how to fix it at the moment.

MichelHollands:
The `TestExpiringCertificate` was a bit flaky. After increasing one timeout there it seems a lot more stable.
Can someone merge this? I don't have permission. The same might apply for making a new release.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	eps := make([]endpoint, diff)
+
+	// The DNS entry for a pod of a stateful set is
+	// ingester-zone-a-0.$(servicename).$(namespace).svc.cluster.local
+	// The service in this case is ingester-zone-a as well.
+	// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
+	for i := 0; i < int(diff); i++ {
+		index := int(*oldReplicas) - i - 1 // nr in statefulset
+		eps[i].url = fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local:%s/ingester/%s",
```
andyasp, line 178: Can `/ingester/` be a part of the path rather than being hardcoded?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## main / unreleased
 
+* [ENHANCEMENT] Add a `prep_downscale` mutating admission webhook that prepares ingesters for shutdown before downscaling. #47
```
andyasp, line 5: Can this be written in a way that doesn't refer to ingesters specifically? I understand that's the use case in mind here, but `rollout-operator` is intentionally generic.

README.md
```diff
@@ -177,22 +177,98 @@ The webhook is conservative and allows changes whenever an error occurs:
 
 Changing the replicas number to `null` (or from `null`) is allowed.
 
+### `/admission/pre-downscale`
+
+This webhook offers a `MutatingAdmissionWebhook` that calls a downscale preparation endpoint on the ingesters for requests that decrease the number of replicas in objects labeled as `grafana.com/prep-downscale: true`.
```
andyasp, line 182: Same ingester reference comment (there is another one further down).

pkg/admission/prep_downscale.go
MichelHollands: Yes, that can be removed.

CHANGELOG.md
MichelHollands: I've changed it to pods. That terminology is used in other parts of the changelog.

README.md
MichelHollands: Both have been changed to pod.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+	port := lbls[PrepDownscalePortKey]
+	if port == "" {
+		reviewResponse := v1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because the %v label is not set or empty.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, PrepDownscalePortKey),
+			},
+		}
+		level.Warn(logger).Log("msg", fmt.Sprintf("downscale not allowed because the %v label is not set or empty", PrepDownscalePortKey))
+		return &reviewResponse
```
colega, line 119: Nit: during this last review, I found it hard to follow the code with all the noise of these denied responses. I would consider not introducing a var for them, and defining a method that would format the `Allowed: false` response, like this:

```suggestion
		level.Warn(logger).Log("msg", fmt.Sprintf("downscale not allowed because the %v label is not set or empty", PrepDownscalePortKey))
		return deny("downscale of %s/%s in %s from %d to %d replicas is not allowed because the %v label is not set or empty.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, PrepDownscalePortKey)

// ...

// deny returns a *v1.AdmissionResponse with Allowed: false and the message provided formatted with as in fmt.Sprintf.
func deny(msg, args ...any) *v1.AdmissionResponse {
	return v1.AdmissionResponse{
		Allowed: false,
		Result: &metav1.Status{
			Message: fmt.Sprintf(msg, args...),
		},
	}
}
```

IMO this would make this entire method shorter and easier to follow.

pkg/admission/prep_downscale.go
```diff
@@ -0,0 +1,244 @@
...
+				Allowed: false,
+				Result: &metav1.Status{
+					Message: fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because finding other statefulsets failed.", ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas),
+				},
+			}
+			level.Warn(logger).Log("msg", "downscale not allowed due to error while adding annotation", "err", err)
+			return &reviewResponse
+		}
+		if foundSts != nil {
+			reviewResponse := v1.AdmissionResponse{
```
colega, line 148: Should we fail if *this* sts was downscaled few moments ago? Can we downscale again the same statefulset? I'm trying to think of corner cases that would bite us here.

pkg/admission/prep_downscale.go
MichelHollands: Good idea. The function was added manually.

pkg/admission/prep_downscale.go
MichelHollands: For Loki we set the `stabilizationWindowSeconds` in the Keda ScaledObject (which is passed to the HPA). This makes sure there is a certain time between downscales. If this is set big enough the other statefulsets should have time to finish their downscales at which point another scale down should be possible.

colega: I was thinking about a human operator having to do something manually. However, a human can just delete the annotation as a break-glass mechanism so it's ok.

colega:
Thank you!

pkg/admission/prep_downscale.go
MichelHollands: @andyasp I just found out this will not work. A label cannot contain `/`. So I'll have to reinstate that `/ingester/` prefix for now.

PR #49, opened 2023-04-20
Update dependencies

andyasp:
Another round of dependency updates.

The following weren't strictly needed (for CVEs), but I pulled them in because we'll have to do a round of testing on this anyway:
- `github.com/prometheus/client_golang`
- `github.com/prometheus/common`
- `go.uber.org/atomic`


pracucci (approved):
I checked the changelogs and I can't see anything critical. LGTM!

PR #51, opened 2023-04-24
Reinstate /ingester/ prefix

MichelHollands:
Labels in Kubernetes do not allow the `/` character. Therefore the `/ingester/` prefix needs to be added.
Found this while working on an integration test for the downscale webhook.


56quarters (approved):

colega:
I don't like having to enforce a given random prefix here. There's no such thing as an ingester in the scope of `rollout-operator`. For example, this feature is very likely going to be used to scale Mimir's store-gateway's too, and I'd feel weird exposing an `/ingester` path on them. 

Discussed on slack with @MichelHollands, it seems that annotations are a better fit to store this data (port & path).

MichelHollands:
Closing this and I'll add another PR with annotations.

PR #52, opened 2023-04-24
Use annotations for the http port and path

MichelHollands:
Previously these used labels which cannot contain the `/` character.

### DIFF
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -256,7 +256,10 @@ func getResourceAnnotations(ctx context.Context, ar v1.AdmissionReview, api kube
 	switch ar.Request.Resource.Resource {
 	case "statefulsets":
 		obj, err := api.AppsV1().StatefulSets(ar.Request.Namespace).Get(ctx, ar.Request.Name, metav1.GetOptions{})
-		return obj.Annotations, err
+		if err != nil {
+				return nil, err
+		}
+		return obj.Annotations, nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/admission/prep_downscale.go
```diff
@@ -234,3 +251,12 @@ func deny(msg string, args ...any) *v1.AdmissionResponse {
 		},
 	}
 }
+
+func getResourceAnnotations(ctx context.Context, ar v1.AdmissionReview, api kubernetes.Interface) (map[string]string, error) {
+	switch ar.Request.Resource.Resource {
+	case "statefulsets":
+		obj, err := api.AppsV1().StatefulSets(ar.Request.Namespace).Get(ctx, ar.Request.Name, metav1.GetOptions{})
+		return obj.Annotations, err
```
aknuds1, line 259: I believe `obj` will be nil if an error is returned:
```suggestion
		if err != nil {
				return nil, err
		}
		return obj.Annotations, nil
```

aknuds1 (changes requested):
I think there's a risk of dereferencing a nil pointer.

aknuds1 (approved):
LGTM

aknuds1 (changes requested):
Could you also add this PR to the following changelog entry, please?

```
* [ENHANCEMENT] Add an `/admission/prepare-downscale` mutating admission webhook that prepares pods for shutdown before downscaling. #47
```

aknuds1 (approved):
LGTM, thanks for applying my suggestions :)

colega (approved):

aknuds1:
I will merge after CI finishes.

PR #53, opened 2023-04-25
WIP Add integration test for downscale

MichelHollands:
Adds an integration test that downscales a statefulset.

Note that the other tests are commented out at the moment and that the cluster is not deleted. This will be reverted in the final version.


JordanRushing (approved):
LGTM! Thanks for implementing this!

PR #62, opened 2023-05-24
Allow statefulsets to be scaled up and down based on each other

56quarters:
Allow statefulsets within a rollout group to declare a "leader" and set their number of replicas based on the number of replicas in the leader statefulset. Scaling up is allowed immediately while scaling down must respect the minimum amount of time between scale downs. For example at minimum, 12 hours needs to pass between scale downs of Mimir ingester zones.


pkg/controller/controller.go
```diff
@@ -198,6 +199,10 @@ func (c *RolloutController) reconcile(ctx context.Context) error {
 	groups := groupStatefulSetsByLabel(sets, RolloutGroupLabel)
 	var reconcileErrs error
 	for groupName, groupSets := range groups {
+		if err := c.reconcileStatefulSetsGroupReplicas(ctx, groupName, groupSets); err != nil {
```
56quarters, line 202: I don't really like the way this functionality integrates with the existing reconcile loop, it feels "bolted on" to me. However, I did it like this to minimize changes to the existing logic and make it easier to develop. WDYT?

README.md
```diff
@@ -27,7 +27,37 @@ For each **rollout group**, the operator **guarantees**:
...
+
+## How scaling up and down works
+
+The operator can also optionally coordinate scaling up and down of `StatefulSets` that are part of the same `rollout-group` based on the `grafana.com/rollout-downscale-leader` annotation. When using this feature, the `grafana.com/min-time-between-zones-downscale` label must also be set on each `StatefulSet`.
+
+This can be useful for automating the tedious scaling of stateful services like Mimir ingesters. Making use of this feature requires adding a few annotations and labels to configure how it works. Examples for a multi-AZ ingester group are given below.
+
+- For `ingester-zone-a`, add the following:
+  - Labels:
+    - `grafana.com/min-time-between-zones-downscale=12h` (change the value here to an appropriate duration)
```
56quarters, line 40: I'd like to make both of these labels into annotations because they don't _need_ to be labels but wanted to get this functionality rolled out first. We'll have to coordinate with Loki for that change.

README.md
```diff
@@ -27,7 +27,37 @@ For each **rollout group**, the operator **guarantees**:
...
-  - `> 1`: pods are rolled out parallelly (honoring the configured number of max unavailable pods)
+  - `> 1`: pods are rolled out in parallel (honoring the configured number of max unavailable pods)
+
+## How scaling up and down works
+
+The operator can also optionally coordinate scaling up and down of `StatefulSets` that are part of the same `rollout-group` based on the `grafana.com/rollout-downscale-leader` annotation. When using this feature, the `grafana.com/min-time-between-zones-downscale` label must also be set on each `StatefulSet`.
+
+This can be useful for automating the tedious scaling of stateful services like Mimir ingesters. Making use of this feature requires adding a few annotations and labels to configure how it works. Examples for a multi-AZ ingester group are given below.
+
+- For `ingester-zone-a`, add the following:
```
56quarters, line 38: In the future, I think we can just make using the rollout-operator to scale ingesters and store-gateways the default in the Mimir Jsonnet / Helm.

pracucci (approved):
Very good job! The code is clean and every easy to follow. I just left 1 major comment about skipping the rest of the reconcile if we do any change to replicas. Other comments are nits / non blocking. Thanks!

pkg/controller/replicas.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/grafana/rollout-operator/pkg/admission"
+)
+
+type ScaleAction int
+
+const (
+	NoChange ScaleAction = iota
+	ScaleUp
+	ScaleDown
+)
```
colega, line 20: Nit: not sure if these are going to stay (see Marco's comment below), but if they do, these don't need to be exported.

colega:
Very clean work! Agree with Marco's comments, especially about changing replicas when reconciling each statefulset group.

pkg/controller/controller.go
56quarters: Makes sense, thank you. I'll refactor a bit to move the replica reconciliation around.

pkg/controller/replicas.go
```diff
@@ -0,0 +1,160 @@
...
+
+type replicaChanges struct {
+	action   ScaleAction
+	replicas int32
+}
+
+func reconcileStsReplicas(group string, sts *v1.StatefulSet, all []*v1.StatefulSet, logger log.Logger) (replicaChanges, error) {
```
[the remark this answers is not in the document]
56quarters: Sure, good idea.

pkg/controller/controller.go
```diff
@@ -213,6 +218,52 @@ func (c *RolloutController) reconcile(ctx context.Context) error {
...
+				"group", groupName,
+				"name", sts.GetName(),
+				"replicas", res.replicas,
+			)
+
+			patch := fmt.Sprintf(`{"spec":{"replicas":%d}}`, res.replicas)
+			_, err = c.kubeClient.AppsV1().StatefulSets(c.namespace).Patch(ctx, sts.GetName(), types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{})
+			// Return early no matter what after scaling down a single statefulset. We do this because we
+			// need to be sure the last-downscaled annotation on each statefulset object is up-to-date when
+			// we look at it to decide if we can scale down.
```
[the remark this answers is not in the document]
56quarters: Fixed in 47883c74c59e28e03da6587b260af0b3cf024b70

pkg/controller/controller_test.go
```diff
@@ -414,6 +414,8 @@ func TestRolloutController_ReconcileShouldDeleteMetricsForDecommissionedRolloutG
 }
 
 func mockStatefulSet(name string, overrides ...func(sts *v1.StatefulSet)) *v1.StatefulSet {
+	replicas := int32(3)
```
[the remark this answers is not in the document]
56quarters: Added two test cases for `TestRolloutController_Reconcile` in 47883c74c59e28e03da6587b260af0b3cf024b70

pkg/controller/replicas.go
56quarters: Removed in 47883c74c59e28e03da6587b260af0b3cf024b70

pracucci (approved):
Great work, LGTM, thanks!

PR #75, opened 2023-08-03
Handle the context in webhook

MichelHollands:
This will stop the request from adding an annotation to the statefulset if the context is cancelled.


pkg/admission/no_downscale.go
```diff
@@ -140,6 +140,12 @@ func allowErr(logger log.Logger, msg string, err error) *v1.AdmissionResponse {
 }
 
 func getResourceLabels(ctx context.Context, ar v1.AdmissionReview, api kubernetes.Interface) (map[string]string, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+
 	switch ar.Request.Resource.Resource {
```
colega, line 149: Why is this needed? All the calls below have the contexts provided.

colega:
It is usually unneeded to add the `ctx.Done()` check in the code that is not doing something that should be immediately interrupted. Why do we need it here? It seems that all kubernetes client calls already have the context provided, so I'd expect them to check it. Don't they?

pkg/admission/no_downscale.go
```diff
@@ -140,6 +140,12 @@ func allowErr(logger log.Logger, msg string, err error) *v1.AdmissionResponse {
 }
 
 func getResourceLabels(ctx context.Context, ar v1.AdmissionReview, api kubernetes.Interface) (map[string]string, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
```
colega, line 147: Also, this is shorter and equivalent:
```suggestion
	if err := ctx.Err(); err != nil {
		return nil, err
	}
```

MichelHollands:
@colega Sorry, I should have left this as a draft PR. We are having a problem with the prepareDownscale webhook on AWS and Azure where it returns `the HPA controller was unable to update the target scale: Operation cannot be fulfilled on statefulsets.apps "ingester -zone-a": the object has been modified; please apply your changes to the latest version and try again`. This was a test to see if somehow multiple hooks updated the same object and caused the error.

PR #77, opened 2023-08-10
Update Go to 1.21

andyasp:
Updates the intermediate build container and CI to run on [Go 1.21](https://go.dev/doc/go1.21)


andyasp:
As a note tested `go.mod` with `go 1.21` just to see if anything blew up and it was fine. I don't see a need to declare that strict minimum now though so only bumped that to `go 1.20`.

charleskorn (approved):
LGTM, thanks Andy

PR #82, opened 2023-09-14
prepare-downscale webhook should check for rollouts

### LINKED ISSUE #79: Guard against a concurrent downscaling and rollout
At present the rollout operator prevents concurrent rollout operations, but doesn’t prevent concurrent downscale + rollout operations. This can happen, for example, if the controller is currently rolling zone B or C, and then something downscales zone A. 

In order to guard against this, I propose using an annotation on the resource being scaled to represent a “lock” across the scaling group. Ex:

```yaml
grafana.com/scaling-lock: rollout
```

## Creating a scaling-lock

When the rollout operator reconciles, it will choose a resource, if any, that already has a scaling lock to reconcile first, allowing reentrancy. If none exist, it will add a scaling-lock annotation to the sts when it begins rolling its pods. A scaling lock should not be needed though if the controller is just adjusting replicas up/down (since last-downscaled will already provide protection here), only when rolling pods.

## Validating the scaling-lock

When the scaling-lock is present on any resource in the rollout-group, the downscale webhook will reject downscale attempts for resources that don't have the scaling lock. For example: if the scaling lock is present on the zone B statefulset, zone A's statefulset cannot be downscaled.

## Removing the scaling-lock

When the rollout-operator reconciles a resource, if it sees no non-ready pods and has nothing to do for a resource, it will clear the scaling-lock for that resource if any is present.

## Other alternatives

While this could also be accomplished by extending the last-updated annotation to be updated by the rollout-operator’s controller, that would block downscales from taking place for some time after rollouts are complete, which is not necessary. We just want to prevent concurrent scaling from occurring.

jhalterman:
This PR updates the prepare-downscale webhook to check for rollouts that are in progress in the rollout-group, by comparing the statefulset's Status.UpdatedReplicas to the number of ready and running pods in the group.

This PR also:
- Moves some of the util functions that were in the controller package to a separate util package for access by webhooks.
- Removes the unused min/max functions, since these are built-ins in go 1.21

Fixes #79

### DIFF
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -164,20 +164,18 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 			)
 		}
 		if foundSts != nil {
-			level.Warn(logger).Log("msg", "downscale not allowed because another statefulset was downscaled recently", "err", err)
-			return deny(
-				"downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
-				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime,
-			)
+			msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
+				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime)
+			level.Warn(logger).Log("msg", msg, "err", err)
+			return deny(msg)
 		}
 
 		foundSts = findStatefulSetWithNonUpdatedReplicas(ctx, api, ar.Request.Namespace, stsList)
 		if foundSts != nil {
-			level.Warn(logger).Log("msg", "downscale not allowed because another statefulset with non-updated replicas")
-			return deny(
-				"downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v has %d non-updated replicas and %d non-ready replicas",
-				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.nonUpdatedReplicas, foundSts.nonReadyReplicas,
-			)
+			msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v has %d non-updated replicas and %d non-ready replicas",
+				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.nonUpdatedReplicas, foundSts.nonReadyReplicas)
+			level.Warn(logger).Log("msg", msg)
+			return deny(msg)
 		}
 	}
 
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/grafana/rollout-operator/pkg/config"
+	"github.com/grafana/rollout-operator/pkg/util"
 )
 
 const (
@@ -169,7 +170,7 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 			)
 		}
 
-		foundSts = findStatefulSetWithNonUpdatedReplicas(stsList)
+		foundSts = findStatefulSetWithNonUpdatedReplicas(ctx, api, ar.Request.Namespace, stsList)
 		if foundSts != nil {
 			level.Warn(logger).Log("msg", "downscale not allowed because another statefulset with non-updated replicas")
 			return deny(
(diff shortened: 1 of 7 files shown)


pkg/admission/prep_downscale.go
```diff
@@ -159,6 +159,22 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
+		foundSts, err = findStatefulSetWithNonUpdatedReplicas(ctx, api, ar.Request.Namespace, rolloutGroup)
+		if err != nil {
+			level.Warn(logger).Log("msg", "downscale not allowed due to error while finding other statefulsets", "err", err)
+			return deny(
+				"downscale of %s/%s in %s from %d to %d replicas is not allowed because finding other statefulsets failed.",
+				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas,
+			)
+		}
+		if foundSts != nil {
+			level.Warn(logger).Log("msg", "downscale not allowed because another statefulset with non-updated replicas", "err", err)
```
56quarters, line 172: `err` is `nil` here

pkg/admission/prep_downscale.go
```diff
@@ -159,6 +159,22 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime,
 			)
 		}
+
+		foundSts, err = findStatefulSetWithNonUpdatedReplicas(ctx, api, ar.Request.Namespace, rolloutGroup)
+		if err != nil {
+			level.Warn(logger).Log("msg", "downscale not allowed due to error while finding other statefulsets", "err", err)
```
56quarters, line 165: Should we log the same message that we're returning to the caller to help debugging?

pkg/admission/prep_downscale.go
```diff
@@ -159,6 +159,22 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime,
 			)
 		}
+
```
56quarters, line 162: Seems like we're fetching the statefulsets part of this rollout group several times as part of this webhook. Would it make sense to fetch all of them once and pass the objects to each function that needs to check a particular thing about them?

pkg/admission/prep_downscale_test.go
```diff
@@ -258,6 +259,44 @@ func testPrepDownscaleWebhook(t *testing.T, oldReplicas, newReplicas int, option
...
+				Replicas:        10,
+				ReadyReplicas:   4,
+				UpdatedReplicas: 6,
+			},
+		},
+	}
+	api := fake.NewSimpleClientset(objects...)
+
+	sts, err := findStatefulSetWithNonUpdatedReplicas(context.Background(), api, namespace, rolloutGroup)
+	assert.Nil(t, err)
```
56quarters, line 295: Probably want `require.Method()` for these since `assert` doesn't stop execution on failure (so you'll get a `nil` pointer dereference)

pkg/admission/prep_downscale.go
```diff
@@ -329,3 +338,35 @@ func findDownscalesDoneMinTimeAgo(ctx context.Context, api kubernetes.Interface,
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, sts := range stsList.Items {
+		status := sts.Status
+		if !(status.Replicas == status.ReadyReplicas && status.ReadyReplicas == status.UpdatedReplicas) {
+			return &statefulSetDownscale{
+				name:               sts.Name,
+				nonUpdatedReplicas: uint(status.Replicas - status.UpdatedReplicas),
```
56quarters, line 355: Seems like this would be deceptive if there are not ready replicas. Should we return both non-ready and non-updated here?

pkg/admission/prep_downscale.go
```diff
@@ -329,3 +338,35 @@ func findDownscalesDoneMinTimeAgo(ctx context.Context, api kubernetes.Interface,
...
+// may be in the process of being rolled.
+func findStatefulSetWithNonUpdatedReplicas(ctx context.Context, api kubernetes.Interface, namespace, rolloutGroup string) (*statefulSetDownscale, error) {
+	stsList, err := findStatefulSetsForRolloutGroup(ctx, api, namespace, rolloutGroup)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, sts := range stsList.Items {
+		status := sts.Status
+		if !(status.Replicas == status.ReadyReplicas && status.ReadyReplicas == status.UpdatedReplicas) {
```
56quarters, line 352: It might be necessary to check the status of each of the individual pods. I'd double check the logic in the controller because I believe it does something like that since the `status` of the sts lags reality sometimes.

jhalterman: Good suggestions

pkg/admission/prep_downscale.go
jhalterman: Good suggestion

56quarters (approved):
LGTM with some minor feedback. Let's see how it works in a dev environment!

jhalterman:
Tested in dev-01 by attempting to manually downscale a statefulset during a rollout:

> kubectl edit sts/ingester-zone-a
error: statefulsets.apps "ingester-zone-a" could not be patched: admission webhook "prepare-downscale-cortex-dev-01.grafana.com" denied the request: downscale of statefulsets/ingester-zone-a in cortex-dev-01 from 6 to 4 replicas is not allowed because statefulset ingester-zone-b has 0 non-updated replicas and 6 non-ready replicas
You can run `kubectl replace -f /var/folders/c6/z230vgsx5rb3hd3vtj2bc1ww0000gn/T/kubectl-edit-1712830600.yaml` to try this update again.

Then a few minutes later, once the pods were ready:

> kubectl edit sts/ingester-zone-a
statefulset.apps/ingester-zone-a edited

PR #90, opened 2023-10-05
Do not exit reconciliation if scaling errors

jhalterman:
Following #82, downscaling attempts are prevented in a zone if some other zone has non-ready pods. One situation where this could happen is if zone-C is rolling around the same time that zone-A starts experiencing evictions. In this case, zone C may also want to downscale some nodes, but is unable to because zone A has non-ready nodes, and since reconciliation exits when downscaling fails, zone C can't finish rolling its pods either.

This change allows reconciliation to continue even if scaling fails, so that updates may proceed.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -242,7 +242,7 @@ func (c *RolloutController) reconcileStatefulSetsGroup(ctx context.Context, grou
 	// up-to-date.
 	if updated, err := c.adjustStatefulSetsGroupReplicas(ctx, groupName, sets); updated || err != nil {
 		if err != nil {
-			return fmt.Errorf("unable to adjust desired replicas of StatefulSet in group %s: %w", groupName, err)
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -242,7 +242,7 @@ func (c *RolloutController) reconcileStatefulSetsGroup(ctx context.Context, grou
 	// up-to-date.
 	if updated, err := c.adjustStatefulSetsGroupReplicas(ctx, groupName, sets); updated || err != nil {
 		if err != nil {
-			return fmt.Errorf("unable to adjust desired replicas of StatefulSet in group %s: %w", groupName, err)
+			level.Warn(c.logger).Log("msg", "unable to adjust desired replicas of StatefulSet", "group", groupName, "err", err)
```
56quarters, line 245: Seems like this still returns early if there's an error scaling the STS. We only want to return if `updated` is true, right?

56quarters (approved):
Seems reasonable to me, let's try it out

PR #92, opened 2023-10-26
Don't end reconciliation early if the StatefulSet patch failed

56quarters:
Fixes an issue where reconciliation aborted even if the scale down of an STS was denied.


aknuds1 (approved):

andyasp (approved):
LGTM, thanks!

PR #96, opened 2023-11-14
Implement zoneTracker for coordinating downscales across statefulsets using a file in object storage for the prepare_downscale admission webhook

JordanRushing:
This PR introduces the idea of a `zoneTracker` using a JSON-encoded file stored in object storage to save timestamps for coordinating downscales between statefulsets with the `prepare-downscale` admission webhook.

The file's contents:
```json
{
  "ingester-zone-a": {
    "lastDownscaled": "2023-12-14T23:37:41Z"
  },
  "ingester-zone-b": {
    "lastDownscaled": "2023-12-14T23:37:41Z"
  },
  "ingester-zone-c": {
    "lastDownscaled": "2023-12-14T23:37:41Z"
  }
}
```

---

When rolling out the `prepare-downscale` admission webhook to support ingester autoscaling in Loki, we encountered differences in behavior w.r.t the `last-downscale` annotation that the rollout operator was applying to ingester statefulsets between the major cloud service provider's managed Kubernetes control planes. 

HPA controllers in EKS/AKS fail to scale the relevant ingester statefulset after a valid AdmissionRequest is approved in the `prepare-downscale` webhook due to resourceVersion mismatch errors related to the `last-downscale` annotation being applied to the ingester statefulsets. We currently believe this is due to differences in managed control plane configurations between the providers but were unable to pinpoint a root cause even after working with the relevant support team for AWS.

---

Initially, we attempted to submit the annotation patch as a JSONPatch within the AdmissionResponse itself to solve this problem but even that patch was not correctly applied to the ingester statefulsets so downscales were not correctly coordinated.

As a workaround, we considered tracking this state in a file instead of a Kubernetes annotation. A draft [PR](https://github.com/grafana/rollout-operator/pull/84) was opened to do this and it worked with supported changes in the Loki API, but with trade-offs from saving this state on local disk.

After further consideration, we decided to store this file in object storage instead of PVs to simplify and fortify this state tracking with the strong read-after-write consistency guarantees that are provided by the three largest cloud provider's object storage services.

This has been successfully tested as a drop-in replacement for existing namespaces that were using the annotation to coordinate downscales prior on AWS/GCP/Azure (it essentially applies two conditionals to the existing code). It would be possible to add an in-memory bucket option to be configured as well if desired.

---

`go mod vendor` & `go mod tidy` were run as well.

---

#### TODO
- [x] Initial zone file creation
- [x] Bucket configuration and support s3/gcs/abs
- [x] More tests


pkg/admission/prep_downscale.go
```diff
@@ -23,24 +23,33 @@ import (
...
-	return prepareDownscale(ctx, logger, ar, api, client)
+
+	var zt *zoneTracker
+	if useZoneTracker {
+		// TODO(jordanrushing): fix bucket creation semantics and wire-in config supporting multiple CSPs
+		bkt := objstore.NewInMemBucket()
+		zt = newZoneTracker(bkt, "testkey")
+	}
+
+	return prepareDownscale(ctx, logger, ar, api, client, zt)
```
cstyan, line 45: Like I mentioned in our 1:1, I think the easiest option for now is to turn the `zoneTracker` param into a string. If it's not nil, require that it be one of `gcs`, `s3`, `whatever azure storage is called`, etc. That's probably easier than trying to wire up a config file, configmap that populates that config file, etc. so that there could be a `zoneTracker` storage config in the file that specifies the storage backend.

pkg/admission/prep_downscale.go
```diff
@@ -155,27 +164,54 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
+					"downscale of %s/%s in %s from %d to %d replicas is not allowed because loading zones failed.",
+					ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas,
+				)
+			}
+			// Check if the zone has been downscaled recently.
+			foundSts, err := zt.findDownscalesDoneMinTimeAgo(stsList, ar.Request.Name)
+			if err != nil {
+				level.Warn(logger).Log("msg", "downscale not allowed due to error while parsing downscale timestamps from the zone file", "err", err)
+				return deny(
+					"downscale of %s/%s in %s from %d to %d replicas is not allowed because parsing parsing downscale timestamps from the zone file failed.",
```
cstyan, line 182: do we need a log line here and the returning of a warning? couldn't it be just returning the result, or an empty result and an error? then we only deal with logging once where ever it is we're  calling `prepareDownscale` if the result is that we're denying the downscale

pkg/admission/prep_downscale.go
```diff
@@ -155,27 +164,54 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
+			if foundSts != nil {
+				msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
+					ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime)
+				level.Warn(logger).Log("msg", msg, "err", err)
+				return deny(msg)
+			}
+		} else {
+			foundSts, err := findDownscalesDoneMinTimeAgo(stsList, ar.Request.Name)
+			if err != nil {
+				level.Warn(logger).Log("msg", "downscale not allowed due to error while parsing downscale annotations", "err", err)
```
cstyan, line 195: same here

cstyan (commented):
Personally I'd look for a way to start breaking up `prepareDownscale` into some more functions. Especially with the amount of nested conditionals, it's getting a bit difficult to read. Maybe everything that happens in the block for if the zoneTracker file is being used can be a separate function?

pkg/admission/prep_downscale.go
```diff
@@ -155,27 +164,54 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
-		}
-		if foundSts != nil {
-			msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
-				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime)
-			level.Warn(logger).Log("msg", msg, "err", err)
-			return deny(msg)
-		}
+		// If using zoneTracker instead of downscale annotations, check the zone file for recent downscaling.
+		// Otherwise, check the downscale annotations.
+		if zt != nil {
```
MichelHollands, line 169: You could make 2 types (one with the zones file and one with the existing logic) here that implement the same interface. That way the code for them is not intermingled as here.

pkg/admission/zone_tracker.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// If the zone file does not exist, populate it with the current time for each zone and upload it to the bucket.
+func (zt *zoneTracker) createInitialZones(ctx context.Context, stsList *appsv1.StatefulSetList) error {
+	zt.mu.Lock()
+	defer zt.mu.Unlock()
+
+	for _, sts := range stsList.Items {
+		if _, ok := zt.zones[sts.Name]; !ok {
+			zt.zones[sts.Name] = time.Now().UTC().Format(time.RFC3339)
```
MichelHollands, line 127: You can assign `time.Now().UTC().Format(time.RFC3339)` to a variable outside the loop so they all have the same initial time.

pkg/admission/prep_downscale.go
```diff
@@ -155,27 +164,54 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
+				)
+			}
+			if foundSts != nil {
+				msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
+					ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime)
+				level.Warn(logger).Log("msg", msg, "err", err)
+				return deny(msg)
+			}
+		} else {
+			foundSts, err := findDownscalesDoneMinTimeAgo(stsList, ar.Request.Name)
```
MichelHollands, line 193: Do we still need the old way? If the initial load of the zones adds the waiting time to the initial times won't that be enough to migrate?

cstyan:
Mentioned this on slack but posting here too to see if anyone else agrees; it might make sense to change the naming a bit? The name `zoneTracker` is a bit ambiguous in this case imo, and I think it doesn't really match what's actually happening right?
 
 What we're actually tracking is more like the time at which a zone was last autoscaled in order to prevent downscaling too frequently.

My first though is that `zoneTracker` implies we're tracking the zones, like which zones are active for use or something like that. `zoneScalingTracker`, `scalingEventTracker`, `downscaleTracker` would all be less ambiguous imo, but I don't feel strongly enough to block you.

JordanRushing:
> Personally I'd look for a way to start breaking up `prepareDownscale` into some more functions. Especially with the amount of nested conditionals, it's getting a bit difficult to read. Maybe everything that happens in the block for if the zoneTracker file is being used can be a separate function?

I think that's valuable feedback and I'd be happy to open a separate PR to simplify the webhook code and break things out into better functions.

pkg/admission/prep_downscale.go
JordanRushing: I ended up adding the necessary config with flags passed to the zoneTracker instantiation before passing `zt` through to `prepareDownscale`

cmd/rollout-operator/main.go
```diff
@@ -87,6 +103,32 @@ func (cfg config) validate() error {
...
+		}
+		if !supported {
+			return fmt.Errorf("the provided object storage provider is not supported (currently supported providers: %s)", strings.Join(supportedProviders, ", "))
+		}
+		if cfg.objectStorageProvider == "azure" && (cfg.accountName == "" || cfg.objectStorageEndpoint == "") {
+			return errors.New("the account name for the Azure object storage provider has not been specified")
+		}
+		if cfg.objectStorageProvider == "s3" && (cfg.objectStorageEndpoint == "" || cfg.objectStorageRegion == "") {
+			return errors.New("the endpoint and region for the S3 object storage provider have not been specified")
+		}
```
jhalterman, line 129: Do we need a similar check for GCS?

pkg/admission/zone_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	buf := &bytes.Buffer{}
+	if err := json.NewEncoder(buf).Encode(zt.zones); err != nil {
+		return err
+	}
+
+	return zt.bkt.Upload(ctx, zt.key, buf)
+}
+
+// lastDownscaled returns the last time the zone was downscaled in UTC in time.RFC3339 format.
+func (zt *zoneTracker) lastDownscaled(ctx context.Context, zone string) (string, error) {
```
jhalterman, line 64: ctx seems unused

pkg/admission/zone_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+			return err
+		}
+	}
+	defer r.Close()
+
+	return json.NewDecoder(r).Decode(&zt.zones)
+}
+
+// saveZones encodes the zone tracker to json and uploads the zone file to the bucket.
+func (zt *zoneTracker) saveZones(ctx context.Context) error {
```
jhalterman, line 54: Maybe add a doc that callers should lock zt.mu before calling

pkg/admission/zone_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/thanos-io/objstore/providers/azure"
+	"github.com/thanos-io/objstore/providers/gcs"
+	"github.com/thanos-io/objstore/providers/s3"
+	"gopkg.in/yaml.v2"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+type zoneTracker struct {
+	mu    sync.Mutex
+	zones map[string]string
```
jhalterman, line 24: Should we store some struct inside the map rather than just a string? With just the string, this is more of a lastDownscaledTracker than a zone tracker, and I'm wondering if we may want to store other things in there in the future, storing a separate JSON object per zone may make migration easier.

56quarters:
> This PR introduces the idea of a `zoneTracker` using a JSON-encoded file stored in object storage to save timestamps for coordinating downscales between statefulsets with the `prepare-downscale` admission webhook.

Has any consideration been given to using a configmap in Kubernetes instead of introducing a dependency on object storage for the rollout-operator?

cmd/rollout-operator/main.go
JordanRushing: GCS only requires the bucket to be specified, which is checked above this conditional

pkg/admission/zone_tracker.go
JordanRushing: Nice find, removing.

pkg/admission/zone_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	data, err := yaml.Marshal(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func createGCSConfigYaml(ctx context.Context, bucket string) ([]byte, error) {
+	cfg := &gcs.Config{
```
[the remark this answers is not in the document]
JordanRushing: Nice find, removing.

pkg/admission/zone_tracker.go
JordanRushing: Adding.

pkg/admission/zone_tracker.go
JordanRushing: This seems like a reasonable improvement and hopefully also resolves the issue mentioned above with `stsList` only applying to a specific rollout group?

JordanRushing:
> > This PR introduces the idea of a `zoneTracker` using a JSON-encoded file stored in object storage to save timestamps for coordinating downscales between statefulsets with the `prepare-downscale` admission webhook.
> 
> Has any consideration been given to using a configmap in Kubernetes instead of introducing a dependency on object storage for the rollout-operator?

Great question @56quarters! Yes, originally I was considering using a configmap after some deliberation with the team but ultimately, as far as I understand, etcd runs as a part of the managed control plane in EKS/AKS and the differences in the managed control planes between providers + poor visibility motivated this PR so I wanted to move this metadata outside of that.

I wonder if it might be better to move the annotation code path to using a configmap rather than the zoneTracker?

Do you have any concerns about introducing an object storage dependency to an optional code path here? The original code path is currently unchanged.

pkg/admission/prep_downscale.go
JordanRushing: I think that's fair, generally I'm just trying to follow the pattern we've implemented elsewhere in this webhook. I don't think the extra logging has any real negative trade-offs as far as I can tell and this way depending on if you're looking at the operator directly or the admission responses you have visibility into the root cause of the failure.

cmd/rollout-operator/main.go
```diff
@@ -87,6 +103,32 @@ func (cfg config) validate() error {
 	if (cfg.kubeAPIURL == "") != (cfg.kubeConfigFile == "") {
 		return errors.New("either configure both Kubernetes API URL and config file or none of them")
 	}
+	if cfg.useZoneTracker {
```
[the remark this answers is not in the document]
JordanRushing: I don't know if I agree here, this validates the required config fields to launch with the zt following the pattern we use to validate the required kube config as well.

pkg/admission/zone_tracker.go
JordanRushing: Any suggestions for what that struct might look like?

jhalterman: Just something like:

```go
type zoneInfo struct {
	lastDownscaled time.Time
}
```

pkg/admission/prep_downscale.go
```diff
@@ -155,27 +165,54 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
...
-		if foundSts != nil {
-			msg := fmt.Sprintf("downscale of %s/%s in %s from %d to %d replicas is not allowed because statefulset %v was downscaled at %v and is labelled to wait %s between zone downscales",
-				ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldReplicas, *newReplicas, foundSts.name, foundSts.lastDownscaleTime, foundSts.waitTime)
-			level.Warn(logger).Log("msg", msg, "err", err)
-			return deny(msg)
-		}
+		// If using zoneTracker instead of downscale annotations, check the zone file for recent downscaling.
+		// Otherwise, check the downscale annotations.
+		if zt != nil {
+			if err := zt.loadZones(ctx, stsList); err != nil {
```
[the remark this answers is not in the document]
JordanRushing: I've added a comment indicating that statefulset names are expected to be unique even across rollout groups.

pkg/admission/zone_tracker.go
JordanRushing: I implemented this! It's working on initial testing. Thanks for the suggestion.

jhalterman (approved):

jhalterman:
Talking to Jordan, he plans to followup with an attempt at accomplishing this via ConfigMap, which sounds good.

JordanRushing:
> Talking to Jordan, he plans to followup with an attempt at accomplishing this via ConfigMap, which sounds good.

I think we can close this PR in favor of https://github.com/grafana/rollout-operator/pull/107!

PR #99, opened 2023-12-12
prepare-downscale: exclude pods in the same zone when finding unavailable pod

dimitarvdimitrov:
Having unavailable or not up to date pods in the same zone we're trying to scale down shouldn't really stop us from downscaling the zone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## main / unreleased
 
+* [ENHANCEMENT] Proceed with prepare-downscale operation even when pods from zone being downscaled are not ready or not up to date. #99
+
 ## v0.9.0
 
 * [ENHANCEMENT] Updated dependencies, including: #93
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -170,7 +170,7 @@ func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 			return deny(msg)
 		}
 
-		foundSts = findStatefulSetWithNonUpdatedReplicas(ctx, api, ar.Request.Namespace, stsList)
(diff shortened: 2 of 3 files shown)


dimitarvdimitrov:
@56quarters can you take a look at this? I'm not sure if there's an edge case that we can hit. 

I checked what would happen if we try to prepare a non-ready pod for downscale - supposedly the request to it [will fail and the whole downscale will be aborted](https://github.com/grafana/rollout-operator/blob/12ae585fc10cadb5148075c5cf36c23fcb1da517/pkg/admission/prep_downscale.go#L212-L242). But in reality this should be rare because we need only the pods being downscaled to be ready - the rest can be non-ready or not up to date.

jhalterman:
I'm not sure if this restriction was intended for downscales, but availability and resharding is the concern that comes to mind for me, more than the request succeeding.

dimitarvdimitrov:
> I'm not sure if this restriction was intended for downscales, but resharding and avoiding gaps in data is the concern that comes to mind for me, more than the request succeeding.

I'm not sure I follow @jhalterman. If we are downscaling zone-X, then it shouldn't matter that there are unavailable pods in zone-X. We avoid gaps by making sure that the other zones have available pods.

And resharding is I think a natural consequence of scaling down. Not sure if we can avoid it.


Did I understand your comment correctly?

jhalterman:
> If we are downscaling zone-X, then it shouldn't matter that there are unavailable pods in zone-X. We avoid gaps by making sure that the other zones have available pods.

That's right - I forgot we have that cross-sts check. This change seems safe to me then.

jhalterman (approved):

pracucci:
I reviewed this change too and I also think this change is good. It's not much different from the same check done in `findDownscalesDoneMinTimeAgo()`, which allows to downscale the same zone even if it was already downscaled less than "min time" ago.

PR #101, opened 2023-12-14
Add tracing for incoming requests

charleskorn:
This PR adds support for tracing incoming HTTP requests for the two webhooks.

It produces traces like the following, with nested spans for each of the HTTP calls:

Screenshot 2024-01-09 at 3 44 52 pm

I've used OpenTracing here over OpenTelemetry in the interests of consistency with Mimir.

I don't love how invasive these changes were - the default OpenTracing instrumentation for `net/http` requires calling `nethttp.TraceRequest()` for each outgoing HTTP call, but any other option would require either creating our own OpenTracing instrumentation for `net/http`, or using OpenTelemetry, which would break with the existing pattern of Mimir. I'm very open to other ideas or suggestions.

I don't believe we can't add tracing for Kubernetes API calls made by the informers used in the reconciliation loop, as these run outside the context of a parent span, so I haven't added tracing to the reconciliation loop beyond a single span.


cmd/rollout-operator/main.go
```diff
@@ -107,6 +109,17 @@ func main() {
 	ready := atomic.NewBool(false)
 	restart := make(chan string)
 
+	name := os.Getenv("JAEGER_SERVICE_NAME")
```
jhalterman, line 112: How do you think this would be used? Worth adding to the README?

charleskorn: I included this for consistency with how Jaeger is configured in Mimir - very happy to remove it if you think it's not necessary.

jhalterman (approved):
LGTM. I assume we can migrate to OpenTelemetry in the future, when our libraries are ready.

cmd/rollout-operator/main.go
```diff
@@ -118,6 +120,17 @@ func main() {
...
+	if name == "" {
+		name = "rollout-operator"
+	}
+
+	if trace, err := tracing.NewFromEnv(name); err != nil {
+		level.Error(logger).Log("msg", "Failed to setup tracing", "err", err.Error())
```
56quarters, line 129: Logging at error during start up but then continuing on seems odd to me. WDYT about logging at warning _or_ aborting if we can't set up tracing?

pkg/admission/no_downscale.go
```diff
@@ -24,18 +26,27 @@ const (
 
 func NoDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionReview, api *kubernetes.Clientset) *v1.AdmissionResponse {
 	logger = log.With(logger, "name", ar.Request.Name, "resource", ar.Request.Resource.Resource, "namespace", ar.Request.Namespace)
+	spanLogger, ctx := spanlogger.New(ctx, logger, "No downscale", tenantResolver)
```
56quarters, line 29:
```suggestion
	spanLogger, ctx := spanlogger.New(ctx, logger, "admission.NoDownscale()", tenantResolver)
```

pkg/admission/no_downscale.go
```diff
@@ -140,6 +152,13 @@ func allowErr(logger log.Logger, msg string, err error) *v1.AdmissionResponse {
 }
 
 func getResourceLabels(ctx context.Context, ar v1.AdmissionReview, api kubernetes.Interface) (map[string]string, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Get resource labels")
```
56quarters, line 155:
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.getResourceLabels()")
```

pkg/admission/prep_downscale.go
```diff
@@ -45,11 +49,19 @@ func PrepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionRev
 }
 
 type httpClient interface {
-	Post(url, contentType string, body io.Reader) (resp *http.Response, err error)
+	Do(req *http.Request) (*http.Response, error)
 }
 
 func prepareDownscale(ctx context.Context, logger log.Logger, ar v1.AdmissionReview, api kubernetes.Interface, client httpClient) *v1.AdmissionResponse {
 	logger = log.With(logger, "name", ar.Request.Name, "resource", ar.Request.Resource.Resource, "namespace", ar.Request.Namespace)
+	spanLogger, ctx := spanlogger.New(ctx, logger, "Prepare downscale", tenantResolver)
```
56quarters, line 57:
```suggestion
	spanLogger, ctx := spanlogger.New(ctx, logger, "admission.prepareDownscale()", tenantResolver)
```

pkg/admission/no_downscale.go
```diff
@@ -24,18 +26,27 @@ const (
...
+
+	spanLogger.SetTag("object.name", ar.Request.Name)
+	spanLogger.SetTag("object.resource", ar.Request.Resource.Resource)
+	spanLogger.SetTag("object.namespace", ar.Request.Namespace)
 
 	oldObj, oldGVK, err := codecs.UniversalDeserializer().Decode(ar.Request.OldObject.Raw, nil, nil)
 	if err != nil {
 		return allowErr(logger, "can't decode old object, allowing the change", err)
 	}
 	logger = log.With(logger, "request_gvk", oldGVK)
```
56quarters, line 41: I appreciate trying to minimize changes but it seems weird to use both `logger` and `spanLogger`. Can we use just the `spanLogger` and not keep `logger` around in these methods?

andyasp (approved):
Thanks!

pkg/admission/prep_downscale.go
```diff
@@ -185,6 +198,12 @@ func deny(msg string, args ...any) *v1.AdmissionResponse {
 }
 
 func getResourceAnnotations(ctx context.Context, ar v1.AdmissionReview, api kubernetes.Interface) (map[string]string, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Get resource annotations")
```
56quarters, line 201:
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.getResourceAnnotations()")
```

pkg/admission/prep_downscale.go
```diff
@@ -185,6 +198,12 @@ func deny(msg string, args ...any) *v1.AdmissionResponse {
...
+	defer span.Finish()
+
+	span.SetTag("namespace", ar.Request.Namespace)
```
56quarters, line 204: We use `object.namespace` above. Should these match?

pkg/admission/prep_downscale.go
```diff
@@ -197,6 +216,12 @@ func getResourceAnnotations(ctx context.Context, ar v1.AdmissionReview, api kube
 }
 
 func addDownscaledAnnotationToStatefulSet(ctx context.Context, api kubernetes.Interface, namespace, stsName string) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Add downscaled annotation to StatefulSet")
```
56quarters, line 219:
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.addDownscaledAnnotationToStatefulSet")
```

pkg/admission/prep_downscale.go
```diff
@@ -262,6 +287,11 @@ func findDownscalesDoneMinTimeAgo(stsList *appsv1.StatefulSetList, excludeStsNam
 //
 // The StatefulSet whose name matches the input excludeStsName is not checked.
 func findStatefulSetWithNonUpdatedReplicas(ctx context.Context, api kubernetes.Interface, namespace string, stsList *appsv1.StatefulSetList, excludeStsName string) (*statefulSetDownscale, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Find StatefulSet with non-updated replicas")
```
56quarters, line 290:
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.findStatefulSetWithNonUpdatedReplicas")
```

pkg/admission/prep_downscale.go
```diff
@@ -423,22 +465,44 @@ func createEndpoints(ar v1.AdmissionReview, oldInfo, newInfo *objectInfo, port,
 }
 
 func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client httpClient, eps []endpoint) error {
+	if len(eps) == 0 {
+		return nil
+	}
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Prepare pods for shutdown")
```
56quarters, line 472:
```suggestion
	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.sendPrepareShutdownRequests")
```

pkg/admission/prep_downscale.go
```diff
@@ -423,22 +465,44 @@ func createEndpoints(ar v1.AdmissionReview, oldInfo, newInfo *objectInfo, port,
...
+	defer span.Finish()
+
 	g, _ := errgroup.WithContext(ctx)
 	for _, ep := range eps {
 		ep := ep // https://golang.org/doc/faq#closures_and_goroutines
 		g.Go(func() error {
-			logger := log.With(logger, "url", ep.url, "index", ep.index)
+			logger, ctx := spanlogger.New(ctx, log.With(logger, "url", ep.url, "index", ep.index), "Prepare pod for shutdown", tenantResolver)
```
56quarters, line 479:
```suggestion
			logger, ctx := spanlogger.New(ctx, log.With(logger, "url", ep.url, "index", ep.index), "admission.PreparePodForShutdown", tenantResolver)
```

56quarters (approved):
LGTM, nothing major just some stylistic stuff.

charleskorn:
> Apologies @charleskorn, I merged too early. I reverted the commit and restored the branch so you could respond to the above in a reopened PR.

No worries, I'll continue in #114.

cmd/rollout-operator/main.go
charleskorn: Mimir does the same thing, but I've changed it in efae75d.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/no_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/no_downscale.go
charleskorn: Addressed in 90caa2a.

pkg/admission/prep_downscale.go
charleskorn: Good catch, fixed in 1c1bffa

PR #104, opened 2023-12-14
Makefile: ensure a Docker multi-platform builder exists before building images

pracucci:
Today I've encountered an issue when building v0.10.0: on my machine I didn't have a multi-platform builder selected. In this PR I propose to create a specific builder for the rollout-operator (with the platforms we need) and then use it. This fixed my issue.


pstibrany:
I am not sure that creating per-project builder is a way to go. Builders are deamons running on your machine, and if they are unused, they just consume resources. Perhaps it would be better to check if default builder supports required platforms, and if not, suggest commands to set up one?

pracucci:
> I am not sure that creating per-project builder is a way to go. Builders are deamons running on your machine, and if they are unused, they just consume resources. Perhaps it would be better to check if default builder supports required platforms, and if not, suggest commands to set up one?

Ok

pracucci:
_Closing for now. Don't have bandwidth to follow up on this._

PR #105, opened 2023-12-14
Fix: do not allow downscale if the operator failed to check whether there are StatefulSets with non-updated replicas

pracucci:
The error returned by `countRunningAndReadyPods()` is swallowed but shouldn't.


jhalterman (approved):

56quarters (approved):
LGTM

PR #108, opened 2023-12-15
Local testing environment

charleskorn:
This PR adds some Kubernetes manifests and a script to deploy a basic test installation of the rollout-operator in a local Kubernetes cluster. It also deploys an instance of Jaeger for testing #101.

More details in [the readme](https://github.com/grafana/rollout-operator/blob/charleskorn/local-dev-env/development/README.md).


flxbk (approved):
Tested locally, works for me. Thanks, very useful!

development/rollout-operator-role.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: rollout-operator-role
+  namespace: rollout-operator-development
+rules:
+  - apiGroups:
+      -
```
christopherzli, line 8: why is this empty?

charleskorn: An empty value means the "core" group (see https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-example for an example)

christopherzli: `error: error validating "/Users/christopher.li/rollout-operator/development/rollout-operator-role.yaml": error validating data: ValidationError(Role.rules[0].apiGroups): unknown object type "nil" in Role.rules[0].apiGroups[0]; if you choose to ignore these errors, turn validation off with --validate=false` when I try to run ./apply.sh

charleskorn: What version of Kubernetes and `kubectl` are you using?

PR #114, opened 2024-01-11
Add tracing for incoming requests

charleskorn:
Continuation of #101, which was merged and then [reverted](https://github.com/grafana/rollout-operator/pull/113) due to some feedback received after merging.


pkg/admission/no_downscale.go
```diff
@@ -24,18 +26,27 @@ const (
...
+
+	spanLogger.SetTag("object.name", ar.Request.Name)
+	spanLogger.SetTag("object.resource", ar.Request.Resource.Resource)
+	spanLogger.SetTag("object.namespace", ar.Request.Namespace)
 
 	oldObj, oldGVK, err := codecs.UniversalDeserializer().Decode(ar.Request.OldObject.Raw, nil, nil)
 	if err != nil {
 		return allowErr(logger, "can't decode old object, allowing the change", err)
 	}
 	logger = log.With(logger, "request_gvk", oldGVK)
```
charleskorn, line 41: @56quarters [commented](https://github.com/grafana/rollout-operator/pull/101#discussion_r1447468161) in #101:

> I appreciate trying to minimize changes but it seems weird to use both `logger` and `spanLogger`. Can we use just the `spanLogger` and not keep `logger` around in these methods?

charleskorn: I kept both to preserve the behaviour of tagging log lines with the attributes added along the way such as `request_gvk`. The trouble comes from `log.With` returning a `log.Logger`, rather than a `spanlogger.SpanLogger` - so as soon as we call `log.With`, we lose the ability to set span tags.

One option would be to add another method to `SpanLogger` that allows us to add a tag and wrap the underlying logger with another attribute, which would tidy this up - what do you think?

56quarters: That sounds reasonable but let's not block this PR on it. Thanks :+1:

56quarters (approved):

andyasp (approved):

pkg/admission/no_downscale.go
charleskorn: https://github.com/grafana/dskit/pull/467 will add the method needed for this to dskit.

charleskorn: ...and https://github.com/grafana/rollout-operator/pull/115 will use that method to tidy this up

Issue #126, opened 2024-01-24
Operator assumes all pods are ready when there are no pods at all


colega:
We're currently checking `hasStatefulSetNotReadyPods` before considering the zone reconciled:

https://github.com/grafana/rollout-operator/blob/0b99175d41e44ae5b3b6f871e37048c50cb42500/pkg/controller/controller.go#L382-L410

However, we've seen a case where all pods of a statefulset were deleted, but no new pods were created yet. This caused the `listNotReadyPodsByStatefulSet` method return no pods at all, as there were no pods in any state.

This caused an outage, as next zone was terminated immediately, before no pods were available in the first one.

I think that the code should check that there are at least as many pods as replicas desired by the statefulset, and then it should check whether some of them are unready.

PR #127, opened 2024-01-24
Consider missing pods as not ready

colega:
If all pods are missing, we can't say there are no "not ready pods".

Fixes: https://github.com/grafana/rollout-operator/issues/126

Signed-off-by: Oleg Zaytsev <mail@olegzaytsev.com>

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -411,7 +411,7 @@ func (c *RolloutController) hasStatefulSetNotReadyPods(sts *v1.StatefulSet) (boo
 		return true, nil
 	}
 
-	notReadyPods := c.listNotReadyPodsByStatefulSet(pods)
+	notReadyPods := notRunningAndReady(pods)
 	if len(notReadyPods) == 0 {
 		return false, nil
 	}
@@ -439,7 +439,7 @@ func (c *RolloutController) listPodsByStatefulSet(sts *v1.StatefulSet) ([]*corev
 	return pods, nil
 }
 
-func (c *RolloutController) listNotReadyPodsByStatefulSet(pods []*corev1.Pod) []*corev1.Pod {
+func notRunningAndReady(pods []*corev1.Pod) []*corev1.Pod {
 	// Build a list of not-ready ones.
 	// We don't pre-allocate this list because most of the time we expect all pods are running and ready.
 	var notReady []*corev1.Pod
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -392,10 +392,26 @@ func (c *RolloutController) hasStatefulSetNotReadyPods(sts *v1.StatefulSet) (boo
 	// a terminating pod (which we don't consider "ready") may have not yet failed the
 	// readiness probe for the consecutive number of times required to switch its status
 	// to not-ready. For this reason, we list all StatefulSet pods and check them one-by-one.
```
pracucci, line 394: [nit] I would move the comment block ☝️ before the following line:

```
notReadyPods := notRunningAndReady(pods)
```

pkg/controller/controller.go
```diff
@@ -392,10 +392,26 @@ func (c *RolloutController) hasStatefulSetNotReadyPods(sts *v1.StatefulSet) (boo
...
+	// Sometimes it takes a while until pods are created after terminating them.
+	// We consider the missing pods as not ready.
+	pods, err := c.listPodsByStatefulSet(sts)
 	if err != nil {
 		return false, err
 	}
+
+	if len(pods) < int(sts.Status.Replicas) {
+		level.Info(c.logger).Log(
+			"msg", "StatefulSet status is reporting all pods ready, but the rollout operator found less pods than expected",
```
pracucci, line 406: > StatefulSet status is reporting all pods ready

I don't think this is true. We haven't checked their readiness yet. I would update the log message to better reflect the check we've done.

pracucci (approved):
Thanks @colega. I think the change is safe and fixes a real issue. I left a couple of nits I would be glad you could take a look at before merging.

pkg/controller/controller.go
colega: The message refers to the fact that if we're here, then `sts.Status.Replicas == sts.Status.ReadyReplicas`.

pkg/controller/controller.go
colega: Makes sense, moved in https://github.com/grafana/rollout-operator/pull/127/commits/85a9bb6227989a71dbf8e8d9ef5a87db32980cbf

pkg/controller/controller.go
pracucci: Yes, but doesn't mean "all pods ready". It's just a wrong message we're sending through the logs.

pracucci: Sorry, I'm wrong. All good here.

pr00se (approved):
Thank you!

PR #129, opened 2024-01-29
Mirror replicas from different resource

pstibrany:
This PR allows the statefulset's desired replicas to follow replicas in different resource (configured using
* `grafana.com/rollout-mirror-replicas-from-resource-name`
* `grafana.com/rollout-mirror-replicas-from-resource-kind`
* `grafana.com/rollout-mirror-replicas-from-resource-api-version`

annotations.

Note that "status" subresource of target resource is used to determine number of replicas. `status.replicas` field of "from-resource" will be kept up-to-date with current number of replicas of statefulset.

This is similar to existing `grafana.com/rollout-downscale-leader` annotation, but there are some differences:

* compared to `grafana.com/rollout-downscale-leader`, `grafana.com/rollout-mirror-replicas-from-resource-*` annotations can point to any kind of resource, including custom resource
* `grafana.com/min-time-between-zones-downscale` is not respected when using `rollout-mirror-replicas-*` annotations

This can be used in combination with HPA, when it's undesireable to set target number of replicas directly on statefulset, because we want to add custom logic to the scaledown (such logic will be added to rollout-operator in the next step, specifically download-delay). In that case, HPA can update different resource (eg. custom resource), and rollout-operator can "mirror" number of replicas from this resource to target statefulset.


pkg/config/config.go
```diff
@@ -28,4 +28,11 @@ const (
 	// RolloutDownscaleLeaderAnnotationKey is the name of the leader statefulset that should be used to determine
 	// the number of replicas in a follower statefulset.
 	RolloutDownscaleLeaderAnnotationKey = "grafana.com/rollout-downscale-leader"
+
+	RolloutMirrorReplicasFromResourceAnnotationKeyPrefix = "grafana.com/rollout-mirror-replicas-from-resource"
```
jhalterman, line 32: This could probably be unexported?

pkg/controller/controller.go
```diff
@@ -314,46 +325,63 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicas(ctx context.Context,
...
 
-			return true, nil
+		direction := ""
+		if desiredReplicas > currentReplicas {
+			direction = "up"
+		} else if desiredReplicas < currentReplicas {
+			direction = "down"
+		}
+
+		level.Info(c.logger).Log("msg", fmt.Sprintf("scaling %s statefulset %s", direction, reason), "group", groupName, "name", sts.GetName(), "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas)
```
jhalterman, line 367: We should only log and patch the statefulset if `direction != ""`

pkg/controller/custom_resource_replicas.go
```diff
@@ -0,0 +1,96 @@
...
+	annotations := sts.GetAnnotations()
+	name := annotations[config.RolloutMirrorReplicasFromResourceNameAnnotationKey]
+	kind := annotations[config.RolloutMirrorReplicasFromResourceKindAnnotationKey]
+	if name == "" || kind == "" {
+		return nil, schema.GroupVersionResource{}, "", nil
+	}
+
+	apiVersion := annotations[config.RolloutMirrorReplicasFromResourceAPIVersionAnnotationKey]
+
+	reference := fmt.Sprintf("%s/%s/%s", kind, sts.Namespace, name)
```
jhalterman, line 31: Maybe move this into the `if` statement where it's used below?

pkg/config/config.go
```diff
@@ -28,4 +28,11 @@ const (
...
+	// RolloutMirrorReplicasFromResourceNameAnnotationKey -- when set (together with "kind" and optionally "api-version" annotations), rollout-operator sets number of
+	// replicas based on replicas in this resource (its scale subresource).
+	RolloutMirrorReplicasFromResourceNameAnnotationKey       = RolloutMirrorReplicasFromResourceAnnotationKeyPrefix + "-name"
+	RolloutMirrorReplicasFromResourceKindAnnotationKey       = RolloutMirrorReplicasFromResourceAnnotationKeyPrefix + "-kind"
```
pracucci, line 36: [non blocking] Have you considered a simpler `grafana.com/rollout-mirror-replicas-from-resource: ResourceTemplate/ingester-zone-a` where kind and name are separated by `/`? `kubectl` supports the same, so should be safe.

cmd/rollout-operator/main.go
```diff
@@ -142,14 +144,34 @@ func main() {
...
+	restMapper, err := apiutil.NewDynamicRESTMapper(kubeConfig, httpClient)
+	check(errors.Wrap(err, "failed to create REST Mapper"))
+
+	// we don't use cached discovery because DiscoveryScaleKindResolver does its own caching,
+	// so we want to re-fetch every time when we actually ask for it
+	scaleKindResolver := scale.NewDiscoveryScaleKindResolver(kubeClient.Discovery())
+	scaleClient, err := scale.NewForConfig(kubeConfig, restMapper, dynamic.LegacyAPIPathResolverFunc, scaleKindResolver)
+	check(errors.Wrap(err, "failed to init scaleClient"))
+
+	dynamic, err := dynamic.NewForConfigAndClient(kubeConfig, httpClient)
```
pracucci, line 167: [nit] Super nit:

```suggestion
	dynamicClient, err := dynamic.NewForConfigAndClient(kubeConfig, httpClient)
```

pkg/controller/controller.go
```diff
@@ -314,46 +325,63 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicas(ctx context.Context,
 	// are up-to-date on the models. If we modify the StatefulSet, they no longer are.
```
pracucci, line 325: Above, the function comment says:

> The method returns "true" early if the number of replicas in any StatefulSet was adjusted.

Maybe we wanted to say "only if the number..."?

pkg/controller/controller.go
```diff
@@ -314,46 +325,63 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicas(ctx context.Context,
...
 	for _, sts := range sets {
 		currentReplicas := *sts.Spec.Replicas
+		reason := "to match leader"
 		desiredReplicas, err := desiredStsReplicas(groupName, sts, sets, c.logger)
 		if err != nil {
 			return false, err
 		}
 
-		if desiredReplicas > currentReplicas {
```
pracucci, line 322: From a design perspective, I’m not much convinced of modifying adjustStatefulSetsGroupReplicas() because it adds complexity mixing 2 different ways of scaling.

I would like to propose you an alternative design that should be easy to do and IMO cleaner:
Rename adjustStatefulSetsGroupReplicas() to adjustStatefulSetsGroupReplicasToFollowLeader()
Add your new logic into a new function adjustStatefulSetsGroupReplicasToMirrorResource()
Then create adjustStatefulSetsGroupReplicas() which calls adjustStatefulSetsGroupReplicasToFollowLeader() first, and if it returns false it will call adjustStatefulSetsGroupReplicasToMirrorResource()

This way we keep the 2 ways of scaling separated.

WDYT?

pkg/controller/custom_resource_replicas.go
pracucci, line 31: [nit] Is this a K8S notation? It's the first time I see it. I'm aware of `kind/name` but not `kind/namespace/name`.

pkg/config/config.go
pstibrany: Sure.

pkg/controller/custom_resource_replicas.go
pstibrany: I've now added use of this into another statement.

pkg/config/config.go
pstibrany: This follows KEDA ScaledObject's `spec.ScaleTargetRef` and HPA's `scaleTargetRef` fields ([doc](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#crossversionobjectreference-v2-autoscaling)), and I copied some code for handling this stuff from HPA codebase. I'd prefer to keep it as-is for now.

pkg/controller/custom_resource_replicas.go
pstibrany: This is only used in error message, and since we expect namespace to be the same as statefulsets we're handling, I've removed it.

pstibrany:
I think this is ready for another review, please take a look.

pkg/config/config.go
pracucci: Yes, sure.

pkg/controller/controller_test.go
```diff
@@ -383,10 +394,106 @@ func TestRolloutController_Reconcile(t *testing.T) {
 			kubePatchErr:        errors.New("cannot patch StatefulSet right now"),
 			expectedDeletedPods: []string{"ingester-zone-b-0", "ingester-zone-b-1"},
 		},
+		"should return early and scale up statefulset based on target custom resource": {
```
pracucci, line 397: [nit] To keep naming consistency

```suggestion
		"should return early and scale up statefulset based on reference custom resource": {
```

pkg/controller/controller_test.go
```diff
@@ -383,10 +394,106 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+					"grafana.com/rollout-mirror-replicas-from-resource-kind":        customResourceGVK.Kind,
+					"grafana.com/rollout-mirror-replicas-from-resource-api-version": customResourceGVK.GroupVersion().String(),
+				})),
+			},
+			customResourceScaleSpecReplicas:   5,
+			customResourceScaleStatusReplicas: 2,
+			expectedPatchedSets:               map[string][]string{"ingester-zone-b": {`{"spec":{"replicas":5}}`}},
+			expectedPatchedResources:          map[string][]string{"my.group/v1/customresources/test/status": {`{"status":{"replicas":5}}`}},
+		},
+		"should return early and scale down statefulset based on target custom resource": {
```
pracucci, line 410: [nit]
```suggestion
		"should return early and scale down statefulset based on reference custom resource": {
```

pkg/controller/custom_resource_replicas.go
```diff
@@ -0,0 +1,150 @@
...
+	if firstErr == nil {
+		firstErr = fmt.Errorf("unrecognized resource")
+	}
+
+	return nil, schema.GroupVersionResource{}, firstErr
+}
+
+// updateStatusReplicasOnMirroredResourceIfNeeded makes sure that scaleObject's status.replicas field is up-to-date.
+// if update fails, error is logged, but not returned to caller.
+func updateStatusReplicasOnMirroredResourceIfNeeded(ctx context.Context, log log.Logger, dynamicClient dynamic.Interface, sts *appsv1.StatefulSet, scaleObj *autoscalingv1.Scale, gvr schema.GroupVersionResource, resName string, replicas int32) {
```
pracucci, line 134: [nit] Should we rename this too?

```suggestion
func updateStatusReplicasOnReferenceResourceIfNeeded(ctx context.Context, log log.Logger, dynamicClient dynamic.Interface, sts *appsv1.StatefulSet, scaleObj *autoscalingv1.Scale, gvr schema.GroupVersionResource, resName string, replicas int32) {
```

pkg/controller/custom_resource_replicas.go
```diff
@@ -0,0 +1,150 @@
...
+			"desiredReplicas", desiredReplicas,
+			"referenceResource", referenceResource,
+		)
+
+		if err := c.patchStatefulSetSpecReplicas(ctx, sts, desiredReplicas); err != nil {
+			return false, err
+		}
+
+		// Make sure that scaleObject's status.replicas field is up-to-date.
+		if scaleObj.Status.Replicas != desiredReplicas {
```
pracucci, line 64: This `if` condition is also inside `updateStatusReplicasOnMirroredResourceIfNeeded()`. I think we can either remove it from here, or remove it from `updateStatusReplicasOnMirroredResourceIfNeeded()` and add this if condition even above where we call `updateStatusReplicasOnMirroredResourceIfNeeded()`.

pracucci (approved):
Thanks for addressing my feedback on the design! It's way more clear and easy to follow now. LGTM!

pkg/controller/custom_resource_replicas.go
pstibrany: Good point. I will keep the condition here to match name of the function ("if needed").

PR #131, opened 2024-02-02
Prepare delayed downscale

pstibrany:
This PR adds support for downscale delay when using [reference resource as a source of number of replicas.](https://github.com/grafana/rollout-operator/pull/129)

Two annotations are used for configuration:
* Delay duration: `grafana.com/rollout-delayed-downscale`
* URL: `grafana.com/rollout-prepare-delayed-downscale-url`.

How it works: When rollout-operator detects that statefulset configured with these annotations should be scaled down, it first calls `POST <URL>`  on all pods that should be scaled down. Pods are expected to prepare for scaledown and return timestamp when such preparation happened (possibly in the past). Any errors from these calls (networking, non-200 status code, invalid JSON object) will cause that downscale is aborted (and retried later, if scaledown condition is still true).

Operator then computes Maximum of returned timestamps. If this maximum is in the past for more than "delay duration", scaledown proceeds. Otherwise, scaledown is not allowed at this point, and will be retried later.

All other pods that are NOT going to be scaled down also receive HTTP request to given URL, however with `DELETE` method. This tells pod to "cancel" any possible preparation for scaledown in the past.

For example, if statefulset has 100 pods, and 30 of them are going to be scaled down, then pods from 0 to 69 receive `DELETE <url>` call, while pods 70 .. 99 receive `POST <url>` call.

If rollout operator detects no change of replicas, or scale up, it will still perform `DELETE <url>` calls, to make sure that any possible previous preparations are cancelled. Since these DELETE calls are repeated all the time, errors from them are ignored.

JSON response expected by rollout-operator is simply `{"timestamp": <unix-seconds as number>}`.

Format of the URL annotation is full URL like `http://pod/prepare-delayed-downscale`, in which the "host" part of the URL will be replaced with pod's fully-qualified domain name, eg. `ingester-zone-b-0.ingester-zone-b.namespace.svc.cluster.local.`.

**How is this different from `grafana.com/min-time-between-zones-downscale`?**

`grafana.com/min-time-between-zones-downscale` is applied to delay downscaling of zones based on downscale timestamp of another zone.

**How is this different from `grafana.com/prepare-downscale-http-path`?**

`grafana.com/prepare-downscale-http-path` together with `grafana.com/prepare-downscale-http-port` annotation are used to signal pod that is being scaled down. Ie. after call to `grafana.com/prepare-downscale-http-path+port`, pod will be downscaled right after.

`grafana.com/rollout-prepare-delayed-downscale-url` tells pod about downscale in the future, which can be minutes or hours later.

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -535,7 +535,7 @@ func TestRolloutController_Reconcile(t *testing.T) {
 			})
 
 			patchedStatefulSets := addPatchStatefulsetReactor(kubeClient, testData.kubePatchErr)
-			scaleClient := createScaleClient(testData.customResourceScaleSpecReplicas, testData.customResourceScaleStatusReplicas, testData.getScaleErr)
+			scaleClient := createFakeScaleClient(testData.customResourceScaleSpecReplicas, testData.customResourceScaleStatusReplicas, testData.getScaleErr)
 			dynamicClient, patchedStatuses := createFakeDynamicClient()
 
 			// Create the controller and start informers.
@@ -617,18 +617,18 @@ func TestRolloutController_ReconcileStatefulsetWithDownscaleDelay(t *testing.T) 
 					withDelayedDownscaleAnnotations(time.Hour, "http://pod/prepare-delayed-downscale")),
 			},
 			httpResponses: map[string]httpResponse{
-				"POST http://ingester-zone-b-4.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale": {statusCode: 200, body: fmt.Sprintf(`{"timestamp": %d}`, now.Add(-70*time.Minute).Unix())},
-				"POST http://ingester-zone-b-3.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale": {statusCode: 200, body: fmt.Sprintf(`{"timestamp": %d}`, now.Add(-70*time.Minute).Unix())},
+				"POST http://ingester-zone-b-4.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale": {statusCode: 200, body: fmt.Sprintf(`{"timestamp": %d}`, now.Add(-70*time.Minute).Unix())},
+				"POST http://ingester-zone-b-3.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale": {statusCode: 200, body: fmt.Sprintf(`{"timestamp": %d}`, now.Add(-70*time.Minute).Unix())},
 			},
 			customResourceScaleSpecReplicas:   3, // We want to downscale to 3 replicas only.
 			customResourceScaleStatusReplicas: 5,
 
 			expectedHttpRequests: []string{
-				"DELETE http://ingester-zone-b-0.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
-				"DELETE http://ingester-zone-b-1.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
-				"DELETE http://ingester-zone-b-2.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
-				"POST http://ingester-zone-b-3.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
-				"POST http://ingester-zone-b-4.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+				"DELETE http://ingester-zone-b-0.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale",
+				"DELETE http://ingester-zone-b-1.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale",
+				"DELETE http://ingester-zone-b-2.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale",
+				"POST http://ingester-zone-b-3.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale",
+				"POST http://ingester-zone-b-4.ingester-zone-b.test.svc.cluster.local./prepare-delayed-downscale",
 			},
 
 			expectedPatchedSets: map[string][]string{"ingester-zone-b": {`{"spec":{"replicas":3}}`}}, // This is the downscale!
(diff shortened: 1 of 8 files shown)


pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+
+	for index := from; index < to; index++ {
+		ep := endpoint{
+			namespace: namespace,
+			podName:   fmt.Sprintf("%v-%v", serviceName, index),
+			index:     index,
+		}
+
+		ep.url = *url
+		ep.url.Host = fmt.Sprintf("%s.%v.%v.svc.cluster.local", ep.podName, serviceName, ep.namespace)
```
pracucci, line 129: [nit] I would use a fully qualified DNS name.

```suggestion
		ep.url.Host = fmt.Sprintf("%s.%v.%v.svc.cluster.local.", ep.podName, serviceName, ep.namespace)
```

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+
+type endpoint struct {
+	namespace string
+	podName   string
+	url       url.URL
+	index     int
+}
+
+// Create prepare-downscale endpoints for pods with index in [from, to) range.
+func createEndpoints(namespace, serviceName string, from, to int, url *url.URL) []endpoint {
```
pracucci, line 113: [nit, non blocking] For clarity?

```suggestion
func createPrepareDownscaleEndpoints(namespace, serviceName string, from, to int, url *url.URL) []endpoint {
```

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+				err := errors.New("HTTP DELETE request returned non-2xx status code")
+				body, readError := io.ReadAll(resp.Body)
+				level.Error(epLogger).Log("msg", "unexpected status code returned when calling DELETE on endpoint", "status", resp.StatusCode, "response_body", string(body))
+				return errors.Join(err, readError)
+			}
+			level.Debug(epLogger).Log("msg", "HTTP DELETE request to endpoint succeeded")
+			return nil
+		})
+	}
+	// We ignore errors, since all errors are already logged, and callers don't need i
```
pracucci, line 250: [nit]

```suggestion
	// We ignore errors, since all errors are already logged, and callers don't need it
```

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+	err := g.Wait()
+	return maxTime, err
+}
+
+func callCancelDelayedDownscale(ctx context.Context, logger log.Logger, client httpClient, endpoints []endpoint) {
+	if len(endpoints) == 0 {
+		return
+	}
+
+	g, _ := errgroup.WithContext(ctx)
```
pracucci, line 218: [question, non blocking] Do we want to `SetLimit()` (even an high value) so that we don't spawn a very high number of concurrent calls in a large Mimir cluster?

pkg/controller/custom_resource_replicas.go
```diff
@@ -37,10 +37,22 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicasToMirrorResource(ctx
...
 			continue
 		}
 
+		// We're going to change number of replicas on the statefulset.
+		// If there is delayed downscale configured on the statefulset, we will first handle delay part, and only if that succeeds,
+		// continue with downscaling or upscaling.
+		if err := checkScalingDelay(ctx, c.logger, sts, client, currentReplicas, desiredReplicas); err != nil {
+			level.Warn(c.logger).Log("msg", "not scaling statefulset due to failed scaling delay check", "group", groupName, "name", sts.GetName(), "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas, "err", err)
+
+			// If delay has not been reached, we can check next statefulset.
```
pracucci, line 51: We can end up in a situation where different zone statefulsets are scaled down at different times. When this happens, what's the impact on the INACTIVE partitions? We're going to lose some replicas for a partition which is still INACTIVE in the ring and potentially queried 🤔 

May the solution is to simply configure a scale down delay > than the ring lookback, so that when replicas being to scale down their partitions aren't queried anymore since "some time".

Makes sense?

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+		// should not happen
+		return nil
+	}
+
+	delay, prepareURL, err := parseDelayedDownscaleAnnotations(sts.GetAnnotations())
+	if delay == 0 || prepareURL == nil || err != nil {
+		return err
+	}
+
+	if desiredReplicas > currentReplicas {
```
pracucci, line 50: [nit] I know we can't reach this function when `desiredReplicas == currentReplicas` but from a logical perspective I think a `>=` is better.

```suggestion
	if desiredReplicas >= currentReplicas {
```

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+
+		eps = append(eps, ep)
+	}
+
+	return eps
+}
+
+func callPrepareDownscaleAndReturnMaxPrepareTimestamp(ctx context.Context, logger log.Logger, client httpClient, endpoints []endpoint) (time.Time, error) {
+	if len(endpoints) == 0 {
+		return time.Now(), nil
```
pracucci, line 139: [nit] I would argue this is an error, cause we don't expect this function to be called with no endpoints.

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+	var (
+		maxTimeMu sync.Mutex
+		maxTime   time.Time
+	)
+
+	type expectedResponse struct {
+		Timestamp int64 `json:"timestamp"`
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
```
pracucci, line 151: [non blocking] Same question about `SetLimit()` here.

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+			if err != nil {
+				level.Error(epLogger).Log("error sending HTTP POST request to endpoint", "err", err)
+				return err
+			}
+
+			defer resp.Body.Close()
+
+			body, readError := io.ReadAll(resp.Body)
+			if readError != nil {
+				level.Error(epLogger).Log("msg", "error reading response from HTTP POST request to endpoint", "err", err)
```
pracucci, line 176:
```suggestion
				level.Error(epLogger).Log("msg", "error reading response from HTTP POST request to endpoint", "err", readErr)
```

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+				return err
+			}
+
+			if resp.StatusCode/100 != 2 {
+				err := errors.New("HTTP DELETE request returned non-2xx status code")
+				level.Error(epLogger).Log("msg", "unexpected status code returned when calling DELETE on endpoint", "status", resp.StatusCode, "response_body", string(body))
+				return errors.Join(err, readError)
```
pracucci, line 183: ` readError` can be never valued here, because we return early.

pkg/controller/delay.go
```diff
@@ -0,0 +1,252 @@
...
+		return fmt.Errorf("failed prepare pods for delayed downscale: %v", err)
+	}
+
+	elapsedSinceMaxTime := time.Since(maxPrepareTime)
+	if elapsedSinceMaxTime < delay {
+		return fmt.Errorf("configured downscale delay %v has not been reached for all pods. elapsed time: %v", delay, elapsedSinceMaxTime)
+	}
+
+	// We can proceed with downscale!
+	level.Info(logger).Log("msg", "downscale delay has been reached on all downscaled pods, proceeding with downscale", "name", sts.GetName(), "delay", delay, "elapsed", elapsedSinceMaxTime)
```
pracucci, line 75: [super nit] Design-wise, it's not this function that proceeds with the downscale, so it's should this function to log it. Maybe we can just remove "proceeding with downscale" from this log?

pkg/controller/custom_resource_replicas.go
pracucci, line 51: I would move this comment before the `continue`. I find it misleading before `updateStatusReplicasOnReferenceResourceIfNeeded()`.

pkg/controller/controller_test.go
```diff
@@ -635,6 +590,321 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+			name := action.(ktesting.PatchAction).GetName()
+			patchedStsNames[name] = append(patchedStsNames[name], string(action.(ktesting.PatchAction).GetPatch()))
+		}
+
+		return false, nil, nil
+	})
+	return patchedStsNames
+}
+
+func createScaleClient(customResourceScaleSpecReplicas int, customResourceScaleStatusReplicas int, getScaleErr error) *fakescale.FakeScaleClient {
```
pracucci, line 877: [nit] To keep it consistent with `createFakeDynamicClient()`

```suggestion
func createFakeScaleClient(customResourceScaleSpecReplicas int, customResourceScaleStatusReplicas int, getScaleErr error) *fakescale.FakeScaleClient {
```

pkg/controller/controller_test.go
```diff
@@ -635,6 +590,321 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+			expectedHttpRequests: []string{
+				"DELETE http://ingester-zone-b-0.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+				"DELETE http://ingester-zone-b-1.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+				"DELETE http://ingester-zone-b-2.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+				"POST http://ingester-zone-b-3.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+				"POST http://ingester-zone-b-4.ingester-zone-b.test.svc.cluster.local/prepare-delayed-downscale",
+			},
+		},
+
+		"scale down is not allowed for zone-a, but is IS allowed for zone-b": {
```
pracucci, line 742:
```suggestion
		"scale down is not allowed for zone-a, but IS allowed for zone-b": {
```

pracucci (approved):
Great work and very nice test! I left few minor comments only. Thanks!

pkg/instrumentation/kubernetes_api_client.go
```diff
@@ -46,7 +50,11 @@ func (k *kubernetesAPIClientInstrumentation) RoundTrip(req *http.Request) (*http
 
 	resp, err := k.next.RoundTrip(req)
 	duration := time.Since(start)
-	instrument.ObserveWithExemplar(req.Context(), k.hist.WithLabelValues(urlToResourceDescription(req.URL.EscapedPath()), req.Method, strconv.Itoa(resp.StatusCode)), duration.Seconds())
+	statusCode := 0
```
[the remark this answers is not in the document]
pstibrany: 👍 Good idea.

pkg/controller/delay.go
pstibrany: I didn't know about `SetLimit`, nice tip! I've set it to 32.

pkg/controller/delay.go
pstibrany: 🤷 It looks as if right-hand (this check) was not coordinated with left hand (previous check) :)

pkg/controller/delay.go
pstibrany: > Maybe we can just remove "proceeding with downscale" from this log?
👍

pstibrany:
Thanks for review, I've applied all your suggestions.

PR #144, opened 2024-05-02
Update dependencies

andyasp:
It has been a couple months since we've last released so this is part of preparing for one.


56quarters (approved):
:upside_down_face: we desperately need to get rid of our dependency on Docker and similar things we use for integration tests.

andyasp:
Agreed, this level of noise isn't sustainable. Doing that might even let us use dependabot/renovate without being spammed.

PR #146, opened 2024-05-09
 Invoke DELETE on pod prepare-downscale path if any POSTs failed

seizethedave:
This addresses a bug in rollout-operator where:

1. Kubernetes receives a request to downscale a statefulset by `X` hosts.
2. The prepare-downscale admission webhook approves the downscale and attempts to prepare `X` pods for shutdown by calling invoking their handler identified by the `grafana.com/prepare-downscale-http-path` and `-port` annotations.
3. At least one of these invocations fails. The admission webhook returns an error to Kubernetes, so the downscale is not approved.
4. 💥 But some hosts may have been prepared for downscale. 💥 

This PR adds cleanup logic to issue `DELETE` requests on all pods if any of the `POST`s failed. Notes:
* `DELETE` calls are attempted once.
* `DELETE` failures are logged but otherwise ignored.
* For simplicity, we'll invoke `DELETE` on all of the pods involved in the scaledown operation, not just ones that received a POST.

This doesn't fix the similar issue where replica count changing from 10->9->10 leaves that one pod prepared for shutdown. (But that's in the works.)

### DIFF
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -159,6 +159,10 @@ func prepareDownscale(ctx context.Context, l log.Logger, ar v1.AdmissionReview, 
 	if err := sendPrepareShutdownRequests(ctx, logger, client, eps); err != nil {
 		// Down-scale operation is disallowed because a pod failed to prepare for shutdown and cannot be deleted
 		level.Error(logger).Log("msg", "downscale not allowed due to error", "err", err)
+
+		// At least one endpoint failed. We're going to "undo" all of these by
+		// sending DELETE requests to all of the potentially-affected pods.
+
 		return deny(
 			"downscale of %s/%s in %s from %d to %d replicas is not allowed because one or more pods failed to prepare for shutdown.",
 			ar.Request.Resource.Resource, ar.Request.Name, ar.Request.Namespace, *oldInfo.replicas, *newInfo.replicas,
@@ -460,6 +464,41 @@ func createEndpoints(ar v1.AdmissionReview, oldInfo, newInfo *objectInfo, port, 
 	return eps
 }
 
+func invokePrepareShutdown(ctx context.Context, method string, parentLogger log.Logger, client httpClient, ep endpoint) error {
+	logger, ctx := spanlogger.New(ctx, parentLogger, "admission.PreparePodForShutdown", tenantResolver)
+	defer logger.Span.Finish()
+
+	logger.SetSpanAndLogTag("url", ep.url)
+	logger.SetSpanAndLogTag("index", ep.index)
+
+	req, err := http.NewRequestWithContext(ctx, method, "http://"+ep.url, nil)
+	if err != nil {
+		level.Error(logger).Log("msg", fmt.Sprintf("error creating HTTP %s request", method), "err", err)
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req, ht := nethttp.TraceRequest(opentracing.GlobalTracer(), req)
+	defer ht.Finish()
+
+	resp, err := client.Do(req)
+	if err != nil {
+		level.Error(logger).Log("msg", fmt.Sprintf("error sending HTTP %s request", method), "err", err)
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode/100 != 2 {
+		err := fmt.Errorf("HTTP %s request returned non-2xx status code", method)
+		body, readError := io.ReadAll(resp.Body)
+		level.Error(logger).Log("msg", "error received from shutdown endpoint", "err", err, "status", resp.StatusCode, "response_body", string(body))
+		return errors.Join(err, readError)
+	}
+	level.Debug(logger).Log("msg", "pod prepare-shutdown handler called", "method", method, "url", ep.url)
+	return nil
+}
+
 func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client httpClient, eps []endpoint) error {
 	if len(eps) == 0 {
 		return nil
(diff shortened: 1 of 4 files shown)


pkg/admission/prep_downscale.go
```diff
@@ -460,6 +464,41 @@ func createEndpoints(ar v1.AdmissionReview, oldInfo, newInfo *objectInfo, port,
...
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req, ht := nethttp.TraceRequest(opentracing.GlobalTracer(), req)
+	defer ht.Finish()
+
+	resp, err := client.Do(req)
+	if err != nil {
+		level.Error(logger).Log("msg", fmt.Sprintf("error sending HTTP %s request", method), "err", err)
+		return err
```
seizethedave, line 487: this line was absent from this routine before. I'm 90% sure that was a bug.

pracucci:
Let's wait for @pstibrany review here. He's the expert in this area.

pr00se (approved):
A couple nits, but otherwise LGTM. Thank you!

I'd also like @pstibrany to take a look to make sure it doesn't conflict with anything he's been working on.

pkg/admission/prep_downscale.go
```diff
@@ -468,44 +513,42 @@ func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client
 	span, ctx := opentracing.StartSpanFromContext(ctx, "admission.sendPrepareShutdownRequests()")
 	defer span.Finish()
 
-	g, _ := errgroup.WithContext(ctx)
-	for _, ep := range eps {
-		ep := ep // https://golang.org/doc/faq#closures_and_goroutines
-		g.Go(func() error {
-			logger, ctx := spanlogger.New(ctx, logger, "admission.PreparePodForShutdown", tenantResolver)
-			defer logger.Span.Finish()
+	// Attemt to POST to every prepare-shutdown endpoint. If any fail, we'll
```
pstibrany, line 516: typo: `Attemt` -> `Attempt`.

pkg/admission/prep_downscale.go
```diff
@@ -468,44 +513,42 @@ func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client
...
-			req, ht := nethttp.TraceRequest(opentracing.GlobalTracer(), req)
-			defer ht.Finish()
-
-			resp, err := client.Do(req)
-			if err != nil {
-				level.Error(logger).Log("msg", "error sending HTTP POST request", "err", err)
+	g, ectx := errgroup.WithContext(ctx)
+	g.SetLimit(maxGoroutines)
+	for _, ep := range eps {
+		ep := ep
```
pstibrany, line 524: nit: we don't need this in go 1.22 anymore (we use go 1.22 in rollout-operator). same comment in next for-loop.

pstibrany (commented):
Looks good, thank you.

I checked how Mimir ingester behaves on DELETE method when using ingest storage, and it simply doesn't support that option (returns error), and will stay "prepared for downscale", which means it will unregister from rings and flush storage. When using ingest storage, prepare-for-shutdown should only be called after period in which ingester was not receiving any data for some time already. If pod comes back, it will register back and activate the partition. Overall, I think this is fine.

pkg/admission/prep_downscale.go
```diff
@@ -152,13 +153,16 @@ func prepareDownscale(ctx context.Context, l log.Logger, ar v1.AdmissionReview,
...
-	// Since it's a downscale, check if the resource has the label that indicates it needs to be prepared to be downscaled.
 	// Create a slice of endpoint addresses for pods to send HTTP POST requests to and to fail if any don't return 200
 	eps := createEndpoints(ar, oldInfo, newInfo, port, path)
 
 	if err := sendPrepareShutdownRequests(ctx, logger, client, eps); err != nil {
 		// Down-scale operation is disallowed because a pod failed to prepare for shutdown and cannot be deleted
 		level.Error(logger).Log("msg", "downscale not allowed due to error", "err", err)
+
+		// At least one endpoint failed. We're going to "undo" all of these by
+		// sending DELETE requests to all of the potentially-affected pods.
```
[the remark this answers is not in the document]
seizethedave: Whoops, this comment was from my initial scratchings.

pkg/admission/prep_downscale.go
```diff
@@ -468,44 +513,42 @@ func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client
...
-		ep := ep // https://golang.org/doc/faq#closures_and_goroutines
-		g.Go(func() error {
-			logger, ctx := spanlogger.New(ctx, logger, "admission.PreparePodForShutdown", tenantResolver)
-			defer logger.Span.Finish()
+	// Attemt to POST to every prepare-shutdown endpoint. If any fail, we'll
+	// need to undo them with a DELETE.
 
-			logger.SetSpanAndLogTag("url", ep.url)
-			logger.SetSpanAndLogTag("index", ep.index)
+	const maxGoroutines = 32
```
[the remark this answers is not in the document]
seizethedave: Nothing magic about 32, just to smooth out potentially spiky network IO.

pkg/admission/prep_downscale.go
seizethedave: This makes me a little paranoid as I know others are using this tool. We specify 1.22 in our go.mod but IIUC you can still build and run with an older compiler. And if you don't run the tests it'll probably be race central. :) 
Unless you know something that can lessen my paranoia, I think I'll leave it.

pkg/admission/prep_downscale.go
seizethedave: 👍  Ah yeah, it was the one after http.NewRequest, not client.Do.

https://github.com/grafana/rollout-operator/blob/ea171936c5266041fda49506bac3aa2aa23e075e/pkg/admission/prep_downscale.go#L481-L484

pkg/admission/prep_downscale_test.go
```diff
@@ -269,6 +277,106 @@ func testPrepDownscaleWebhook(t *testing.T, oldReplicas, newReplicas int, option
...
+			}
+
+			err := sendPrepareShutdownRequests(context.Background(), log.NewNopLogger(), httpClient, endpoints)
+			if c.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			if c.lastPostsFail > 0 {
+				assert.Greater(t, postCalls.Load(), int32(0))
```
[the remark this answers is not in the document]
seizethedave: I went back in and tightened this up a little more, as we can expect at least `endpoints-failures` POSTs. But yep you are right about why we can't do an equality check.

pkg/admission/prep_downscale.go
andyasp: > We specify 1.22 in our go.mod but IIUC you can still build and run with an older compiler.

It used to be like that, but now it's mandatory. From https://go.dev/doc/modules/gomod-ref#go
> The go directive sets the minimum version of Go required to use this module. Before Go 1.21, the directive was advisory only; now it is a mandatory requirement: Go toolchains refuse to use modules declaring newer Go versions.

PR #149, opened 2024-05-30
Swap base image from alpine to distroless

andyasp:
Mimir is moving to a distroless image so it would be nice if rollout-operator did too. The primary motivation is to lessen CVE noise even further.

~Opening this as a draft at first to experiment.~ Tests look good!


johannaratliff (approved):
LGTM

PR #151, opened 2024-05-31
Admission webhook: Undo prepare-shutdown calls if last-downscale fails to store

seizethedave:
Fix a snag found in #146 where if the "downscaled" annotation/configmap fails to persist, the scale operation is denied, but the pods are not informed via `DELETE` that they should no longer shutdown.

### DIFF
--- a/pkg/admission/prep_downscale_test.go
+++ b/pkg/admission/prep_downscale_test.go
@@ -61,8 +61,8 @@ func TestValidDownscaleWithAnotherInProgress(t *testing.T) {
 }
 
 func TestInvalidDownscale(t *testing.T) {
-	testPrepDownscaleWebhook(t, 5, 3, withStatusCode(http.StatusInternalServerError), expectDeletes())
-	testPrepDownscaleWebhookWithZoneTracker(t, 5, 3, withStatusCode(http.StatusInternalServerError), expectDeletes())
+	testPrepDownscaleWebhook(t, 5, 3, withStatusCode(http.StatusInternalServerError), expectDeletes(), expectDeny())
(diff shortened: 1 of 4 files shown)


pkg/admission/prep_downscale_test.go
```diff
@@ -58,8 +61,13 @@ func TestValidDownscaleWithAnotherInProgress(t *testing.T) {
 }
 
 func TestInvalidDownscale(t *testing.T) {
-	testPrepDownscaleWebhook(t, 5, 3, withStatusCode(http.StatusInternalServerError))
-	testPrepDownscaleWebhookWithZoneTracker(t, 5, 3, withStatusCode(http.StatusInternalServerError))
+	testPrepDownscaleWebhook(t, 5, 3, withStatusCode(http.StatusInternalServerError), expectDeletes())
```
pr00se, line 64: should these `expectDeny()` as well?

pkg/admission/prep_downscale_test.go
```diff
@@ -275,53 +331,48 @@ func testPrepDownscaleWebhook(t *testing.T, oldReplicas, newReplicas int, option
...
-			expectDeletes: 64,
 			expectErr:     true,
 		},
 		"last post fails": {
 			numEndpoints:  64,
 			lastPostsFail: 1,
-			expectDeletes: 64,
 			expectErr:     true,
 		},
 		"delete failures should still call all deletes": {
```
pr00se, line 366: is this test now identical to the "all posts fail" one above?

pr00se (commented):
A couple questions, but overall LGTM

seizethedave:
Thanks for the read.

pr00se (approved):
LGTM, thanks!

PR #152, opened 2024-06-05
Include username in 'handling request' log.

seizethedave:
Who is calling the webhook? Often this would be useful to know. Sometimes it is rollout-operator itself patching a resource.


56quarters (approved):
LGTM

PR #153, opened 2024-06-10
Support percentages in rollout-max-unavailable annotation

pstibrany:
This PR enhances `rollout-max-unavailable` annotation to support values specified as percentage of statefulset replicas. Eg. `rollout-max-unavailable: 50%`. Returned value is `floor(replicas * percentage)`, but never less than 1.


duricanikolic (approved):
LGTM

Issue #155, opened 2024-06-14
rollout-operator: faster scale down of prepared ingesters


pstibrany:
Rollout-operator currently allows "delayed downscale" if delay has elapsed for all downscaled replicas. For example for following ingest-storage partitions:

<img alt="image" width="983" src="https://private-user-images.githubusercontent.com/895919/328456125-ce7440f9-9a54-42cb-bfdc-36eabd6a6c80.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTgzNjAwMjYsIm5iZiI6MTcxODM1OTcyNiwicGF0aCI6Ii84OTU5MTkvMzI4NDU2MTI1LWNlNzQ0MGY5LTlhNTQtNDJjYi1iZmRjLTM2ZWFiZDZhNmM4MC5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjE0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYxNFQxMDA4NDZaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00NjNiODZiZjU2OTJiZDVjMzQ0MmY0ZDBiZTUyYzA3OWRjMGIxNzk1M2VmMDM2YzkyM2ZjMDU0ZmJmYmU5YzEyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.PDGlF-xddlgi8HcJzLiLj7lHUgQ4sqaJm5qDRLZZiFQ">
maximum timestamp when downscale was initiated is `2024-05-07 04:10:00 UTC`, and that tells rollout-operator to wait for delay since this timestamp -- with 13h delay, that would be `2024-05-07 17:10:00 UTC`.

[...]

PR #156, opened 2024-06-14
Allow delayed downscale of subset of pods

pstibrany:
This PR implements earlier delayed downscale of pods that have already reached their delay, even if not ALL pods have reached it yet. 

Fixes https://github.com/grafana/rollout-operator/issues/155

### DIFF
--- a/pkg/controller/custom_resource_replicas.go
+++ b/pkg/controller/custom_resource_replicas.go
@@ -45,7 +45,7 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicasToMirrorResource(ctx 
 		// We're going to change number of replicas on the statefulset.
 		// If there is delayed downscale configured on the statefulset, we will first handle delay part, and only if that succeeds,
 		// continue with downscaling or upscaling.
-		desiredReplicas := referenceResourceDesiredReplicas
+		var desiredReplicas int32
 		if newDesiredReplicas, err := checkScalingDelay(ctx, c.logger, sts, client, currentReplicas, referenceResourceDesiredReplicas); err != nil {
 			level.Warn(c.logger).Log("msg", "not scaling statefulset due to failed scaling delay check",
 				"group", groupName,
--- a/pkg/controller/custom_resource_replicas.go
+++ b/pkg/controller/custom_resource_replicas.go
@@ -34,10 +34,10 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicasToMirrorResource(ctx 
 
 		referenceResource := fmt.Sprintf("%s/%s", referenceGVR.Resource, referenceName)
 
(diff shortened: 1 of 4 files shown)


pkg/controller/delay.go
```diff
@@ -36,21 +36,24 @@ func cancelDelayedDownscaleIfConfigured(ctx context.Context, logger log.Logger,
...
 	if currentReplicas == desiredReplicas {
 		// should not happen
-		return nil
+		return currentReplicas, nil
 	}
 
 	delay, prepareURL, err := parseDelayedDownscaleAnnotations(sts.GetAnnotations())
 	if delay == 0 || prepareURL == nil || err != nil {
-		return err
+		return desiredReplicas, err
```
pracucci, line 50: If there was an error would be more correct to return `currentReplicas`. Today the caller doesn't scale to `updatedDesiredReplicas` in case of error, but I would like to protect from future bugs.

pkg/controller/custom_resource_replicas.go
```diff
@@ -34,37 +34,47 @@ func (c *RolloutController) adjustStatefulSetsGroupReplicasToMirrorResource(ctx
...
 			// No change in the number of replicas: don't log because this will be the result most of the time.
 			continue
 		}
 
 		// We're going to change number of replicas on the statefulset.
 		// If there is delayed downscale configured on the statefulset, we will first handle delay part, and only if that succeeds,
 		// continue with downscaling or upscaling.
-		if err := checkScalingDelay(ctx, c.logger, sts, client, currentReplicas, desiredReplicas); err != nil {
-			level.Warn(c.logger).Log("msg", "not scaling statefulset due to failed scaling delay check", "group", groupName, "name", sts.GetName(), "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas, "err", err)
+		var desiredReplicas int32
```
pracucci, line 48: To make code more robust from future changes, I suggest to not have this initialised to 0. We could simply do:

```suggestion
desiredReplicas, err := checkScalingDelay(ctx, c.logger, sts, client, currentReplicas, referenceResourceDesiredReplicas)
if err != nil {
   // ...
}
```

pracucci (approved):
Nice job Peter! LGTM. I left a couple of minor comments I would be glad if you could look at before merging. Thanks!

pkg/controller/delay.go
pracucci: >And follow-up fix in https://github.com/grafana/rollout-operator/commit/7e79c3b05f6c7a93442eba9bee65d17b4720cebd.

Exactly. That's what I had in mind.

pstibrany: Fortunately test caught my mistake :)

PR #162, opened 2024-07-15
Allow for podSelectorKey

aallawala:
This PR allows for the Grafana Rollout Operator to customize the key used to select for pods. My org's internal infra uses a different label from `name` and customizing it is trickier than having the rollout operator use for a different key instead.


aallawala:
I see that https://github.com/grafana/rollout-operator/pull/18 is currently pending review. This would also solve our problem, if it is a better solution than this.

Issue #125, opened 2024-01-24
Operator calling the wrong pod endpoint when trying to scale down ingesters


grecuionut:
We are trying to scale down ingesters using the MutatingAdmissionWebhook  as described [here](https://github.com/grafana/rollout-operator?tab=readme-ov-file#admissionprepare-downscale).

The required labels and annotations were added to the objects as shown below:
```
# labels
grafana.com/prepare-downscale=true

# annotations
grafana.com/prepare-downscale-http-path=ingester/prepare-shutdown
grafana.com/prepare-downscale-http-port=8080
```

When trying the scale down the `statefulset/mimir-ingester-zone-a`, the operator failing to resolve the pod when sending HTTP post request, as the fqdn is constructed as `<pod_name>.<service_name>.<namespace>.svc.cluster.local`.
```
mimir-ingester-zone-a-1.mimir-ingester-zone-a.mimir.svc.cluster.local
```

These are the existing services:
```
mimir-ingester-headless                  ClusterIP   None             <none>        8080/TCP,9095/TCP   19d
mimir-ingester-zone-a                    ClusterIP   <ip_address>     <none>        8080/TCP,9095/TCP   19d
mimir-ingester-zone-b                    ClusterIP   <ip_address>    <none>        8080/TCP,9095/TCP   19d
mimir-ingester-zone-c                    ClusterIP   <ip_address>    <none>        8080/TCP,9095/TCP   19d
```

In order to resolve the pod, the headless service should be used instead (mimir-ingester-headless). [More info](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id)

Operator logs
```
level=error ts=2024-01-16T13:29:51.838387358Z name=mimir-ingester-zone-a resource=statefulsets namespace=mimir request_gvk="autoscaling/v1, Kind=Scale" old_replicas=2 new_replicas=1 url=mimir-ingester-zone-a-1.mimir-ingester-zone-a.mimir.svc.cluster.local:443/ingester/prepare-shutdown index=1 msg="error sending HTTP post request" err="Post \"http://mimir-ingester-zone-a-1.mimir-ingester-zone-a.mimir.svc.cluster.local:443/ingester/prepare-shutdown\": dial tcp: lookup mimir-ingester-zone-a-1.mimir-ingester-zone-a.mimir.svc.cluster.local on <name_server_ip>:53: no such host"
```

jhychan:
[...]

PR #173, opened 2024-09-26
separate statefulset name and statefulset service name in downscaling

fitz7:
I found that the rollout-operator was not calling the right pod endpoint with our mimir installation but was calling `mimir-ingester-zone-c-1.mimir-ingester-zone-c.mimir.svc.cluster.local` when it should be using the headless service at `mimir-ingester-zone-c-1.mimir-ingester-headless.mimir.svc.cluster.local`

This led me to issue #125 so checked on our statefulsets and found the serviceName was correctly set to `serviceName: mimir-ingester-headless`

so I dug around a bit and found that the operator was using the an arg named serviceName in `createPrepareDownscaleEndpoints` but passing it the statefulset name.

So I've split them to use both the statefulsets name and it's serviceName so that hopefully these requests go to the correct pods

Let me know if this needs changing or if there is another approach that I've missed?


andyasp:
Thanks for the submission! I think there were a few issues here, but due to the delay in our response I've reimplemented a fix in https://github.com/grafana/rollout-operator/pull/221.

PR #185, opened 2024-12-10
Add duration to `callCancelDelayedDownscale` logs

dimitarvdimitrov:
I'm debugging a reconcilliatin which took 20m. I suspect one ingester was blocking the whole reconcile loop.


pracucci (approved):
Nice approach. Never thought about using `log.Valuer`.

PR #186, opened 2024-12-10
Set timeout for k8s HTTP client

dimitarvdimitrov:
We had a request to an ingesters take 20m and block the reconiliation loop. We want to avoid that, so adding a timeout. We use this client for both k8s and StatefulSet requests.

__I assume the operations that the StatefulSet requests will always be fast and will never have to actually do a lot of work (like uploading blocks). Can the reviewers confirm this assumption?__


pracucci (approved):
I don't expect any K8S API call to take more than few seconds by design (I mean, unless issues with the controlplane). So this PR LGTM.

PR #190, opened 2025-01-03
Introduce a secondary grouping key for rollout operation

boldfield:
The way rollout-operator currently is designed in percludes deploying all of the LGTM stack to a single namespace (which is required in some circumstances). This adds an optional additional label that can be used to employ a secondary grouping to allow for all components to be deployed in a single namespace and have the rollout-operator execute as expected. For example:

Loki ingester:
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    name: loki-ingester-zone-a
    rollout-group: ingester
    rollout-secondary-group: loki
```

Mimir ingester:
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    name: mimir-ingester-zone-a
    rollout-group: ingester
    rollout-secondary-group: mimir
```

### DIFF
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -117,8 +117,9 @@ func prepareDownscale(ctx context.Context, l log.Logger, ar v1.AdmissionReview, 
 	}
 
 	rolloutGroup := lbls[config.RolloutGroupLabelKey]
+	rolloutSecondaryGroup := lbls[config.RolloutSecondaryGroupLabelKey]
 	if rolloutGroup != "" {
-		stsList, err := findStatefulSetsForRolloutGroup(ctx, api, ar.Request.Namespace, rolloutGroup)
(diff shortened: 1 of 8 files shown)


56quarters:
Thanks for this PR but unfortunately we (Mimir team) don't support or test running Mimir in the same namespace as Loki or Tempo. I think we're pretty unlikely to want to support this. If you really need to run the databases in the same namespace it may be enough to change the `rollout-group` of each database to make them unique (e.g. `mimir-ingester`, `loki-ingester`, etc). Can you expand a bit on why this is required?

boldfield:
@56quarters My team runs in a restricted environment where we are required to deploy all of our k8s workloads in a single namespace. We have been using the approach you suggested with customized helm charts for loki and tempo and have PRs up for both. The tempo PR was merged (https://github.com/grafana/loki/pull/15063) but I'm getting pushback on the Loki PR (https://github.com/grafana/helm-charts/pull/3443) and devised this as an alternative approach which would negate the requirement for a helm chart change.

boldfield:
we were previously using kustomize but unfortunately that has become a non-option in our environment as well 👎 

Thanks for your consideration.

PR #195, opened 2025-02-12
WIP: Add cross-cluster traffic protection

aknuds1:


pkg/middleware/clustervalidation.go
dimitarvdimitrov, line 1: clustervalidation -> cluster_validation judging by how more of the files are called in this repo

pkg/admission/prep_downscale.go
```diff
@@ -35,10 +37,15 @@ const (
 	maxPrepareGoroutines        = 32
 )
 
-func PrepareDownscale(ctx context.Context, logger log.Logger, ar admissionv1.AdmissionReview, api *kubernetes.Clientset, useZoneTracker bool, zoneTrackerConfigMapName string) *admissionv1.AdmissionResponse {
+func PrepareDownscale(ctx context.Context, namespace string, logger log.Logger, ar admissionv1.AdmissionReview, api *kubernetes.Clientset, useZoneTracker bool, zoneTrackerConfigMapName string, metrics *metrics.Metrics) *admissionv1.AdmissionResponse {
+	rt := http.DefaultTransport
+	if namespace != "" {
+		// HTTP client side cluster validation.
+		rt = middleware.ClusterValidationRoundTripper(rt, namespace, logger, metrics)
```
dimitarvdimitrov, line 44: cluster protection seems a bit out of scope for the admission webhooks.

Can we instead inject a roundtripper? That way we can keep the cluster_verification roundtripper only in main.go and not have to create a special package for it either.

dimitarvdimitrov: i think metrics wouldn't need to be exported either if we don't need to instantiate the middleware outside of cmd/rollout-operator

dimitarvdimitrov (commented):
some comments on package structure

pkg/admission/prep_downscale.go
aknuds1: PTAL at my latest revision.

cmd/rollout-operator/main.go
```diff
@@ -38,16 +41,23 @@ import (
 
 const defaultServerSelfSignedCertExpiration = model.Duration(365 * 24 * time.Hour)
 
+var (
+	defaultExcludePaths = []string{"admission/no-downscale", "admission/prepare-downscale"}
```
dimitarvdimitrov, line 45:
```suggestion
	defaultClusterValidationExcludePaths = []string{"admission/no-downscale", "admission/prepare-downscale"}
```

cmd/rollout-operator/main.go
```diff
@@ -152,6 +170,20 @@ func main() {
...
+		// HTTP client side cluster validation.
+		reporter := func(msg string, method string) {
+			level.Warn(logger).Log("msg", msg, "method", method, "cluster_validation_label", cfg.kubeNamespace)
+			metrics.InvalidClusterValidationLabels.WithLabelValues(method, "http", cfg.kubeNamespace).Inc()
+		}
+		httpRT = middleware.ClusterValidationRoundTripper(cfg.kubeNamespace, reporter, httpRT)
+
+		kubeConfig.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+			return httpRT
+		})
```
dimitarvdimitrov, line 184: shouldn't we wrap the `rt` we get? why are we discarding it?

aknuds1: Oops, good catch!

aknuds1: Fixed.

dimitarvdimitrov (approved):
looks good

maybe one last thing is to add some sort of tests to assert that the header is present on outgoing requests. I'm not sure how doable this is, so if you decide, you can also merge as-is

PR #196, opened 2025-02-12
Update Go to 1.24

andyasp:
Release notes: https://go.dev/doc/go1.24

Updated `golangci-lint` version used in the action because `1.64.2` supports 1.24 ([release](https://github.com/golangci/golangci-lint/releases/tag/v1.64.2))

Moved `fmt.Sprintf` calls outside of the `deny` function due to an added check in 1.24:
> The existing printf analyzer now reports a diagnostic for calls of the form fmt.Printf(s), where s is a non-constant format string, with no other arguments. Such calls are nearly always a mistake as the value of s may contain the % symbol; use fmt.Print instead.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: '1.23'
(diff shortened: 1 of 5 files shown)


56quarters (approved):
LGTM

PR #200, opened 2025-03-24
Upgrade dependencies to mitigate security vulnerabilities

emadolsky:
This PR upgrades packages to recent versions. The main reason is to fix security vulnerabilites.


andyasp:
Appreciate it. I assume you have good intentions here, but large changes like this are a security concern on their own. One way for me to verify this would be to replicate the update myself, diff the changes, then review it like normal, but at that point there's no benefit.

The dependencies will be updated before we release another version.

PR #208, opened 2025-04-07
Update golangci-lint to v2

andyasp:
Updating golangci-lint from v1 to v2 requires a migration according to this [documentation](https://golangci-lint.run/product/migration-guide/).

What I did here:
-  Ran `golangci-lint migrate -c .golangci.yaml` locally
-  Manually updated the action since it contains v2 support ([release notes](https://github.com/golangci/golangci-lint-action/releases/tag/v7.0.0))
- Ran `golangci-lint run --fix ./...` and manually fixed what remained
- Cleaned up `.golangci.yaml` to remove settings that did not apply

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -8,16 +8,11 @@ linters:
       exclude-functions:
         - (github.com/go-kit/log.Logger).Log
   exclusions:
-    generated: lax
     presets:
       - comments
       - common-false-positives
       - legacy
       - std-error-handling
-    paths:
-      - third_party$
-      - builtin$
-      - examples$
 formatters:
   enable:
     - gci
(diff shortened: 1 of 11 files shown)


.golangci.yaml
```diff
@@ -1,24 +1,26 @@
...
+      - legacy
+      - std-error-handling
+formatters:
   enable:
     - gci
     - gofmt
-
-linters-settings:
-  errcheck:
-    # Use exclude-functions instead of exclude, since the GitHub Action's caching is broken.
```
aknuds1, line 12: Should we keep this comment?

aknuds1 (approved):
LGTM, though please see my question.

.golangci.yaml
andyasp: Sure. The migration command doesn't preserve comments and I wasn't sure if this was still relevant or not. If you're asking about it that's reason enough for me.

PR #218, opened 2025-04-28
Address zizmor scan results

andyasp:
There were only warnings

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -5,6 +5,9 @@ on:
     - cron: "22 */12 * * *"
   # Allow manually triggering
   workflow_dispatch:
+
+permissions: {}
+
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/renovate.yml
```diff
@@ -5,6 +5,9 @@ on:
     - cron: "22 */12 * * *"
   # Allow manually triggering
   workflow_dispatch:
+
+permissions: {}
```
alexweav, line 9: It does a `Checkout Code` step later on - does it need `contents: read`? :thinking:

alexweav (approved):
LGTM, but please check my question

.github/workflows/renovate.yml
andyasp: The permissions are on the job in line 14.

PR #221, opened 2025-05-08
Use serviceName to construct endpoint for prepare-downscale

andyasp:
Addresses #125.

Relevant Kubernetes documentation here: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
> A StatefulSet can use a [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) to control the domain of its Pods. The domain managed by this Service takes the form: $(service name).$(namespace).svc.cluster.local, where "cluster.local" is the cluster domain. As each Pod is created, it gets a matching DNS subdomain, taking the form: $(podname).$(governing service domain), where the governing service is defined by the serviceName field on the StatefulSet.

When constructing the endpoint to call for a pod's `prepare-downscale`, `rollout-operator` was assuming the `.spec.serviceName` of a `StatefulSet` was equal to the `.metadata.name` of the `StatefulSet`. This happens to be true for each ingester `StatefulSet` in Mimir's Jsonnet, but is not true for ingesters in Mimir's Helm since it has a single headless ingester service.

A community member kindly submitted a fix for this issue in #173, but it never got a response. From inspection I think that implementation missed support for `statefulsets/scale`. It also added another `StatefulSet` query. I implemented these changes from scratch to address these issues and avoid other unnecessary changes.

~Marking as draft until I validate some more.~ Validated locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## main / unreleased
 
 * [CHANGE] Rename metric `rollout_operator_request_invalid_cluster_validation_labels_total` to `rollout_operator_client_invalid_cluster_validation_label_requests_total`. #217
+* [BUGFIX] Use a StatefulSet's `.spec.serviceName` when constructing the prepare-downscale endpoint for a pod. #221
 
 ## v0.26.0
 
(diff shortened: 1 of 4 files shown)


aknuds1 (approved):
LGTM

PR #234, opened 2025-06-18
Migrate tracing to OpenTelemetry and add support for standard OTel env config

colega:
This migrates this app to use OpenTelemetry tracing library instead of OpenTracing, which has been deprecated for a while.

The app will now support standard 'OTEL_' environment variables as per OTel SDK docs, and the exposition in OTLP protocol.

### DIFF
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -534,7 +534,7 @@ func invokePrepareShutdown(ctx context.Context, method string, parentLogger log.
 }
 
 func sendPrepareShutdownRequests(ctx context.Context, logger log.Logger, client httpClient, eps []endpoint) error {
-	ctx, span := tracer.Start(ctx, "admission.sendPrepareShutdownRequests()", trace.WithAttributes())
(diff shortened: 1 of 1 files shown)


development/rollout-operator-deployment.yaml
```diff
@@ -41,16 +41,6 @@ spec:
...
-            - name: JAEGER_SAMPLER_TYPE
-              value: const
-            - name: JAEGER_SAMPLER_PARAM
-              value: "1"
-            - name: JAEGER_TAGS
-              value: app=rollout-operator
-            - name: JAEGER_REPORTER_MAX_QUEUE_SIZE
-              value: "1000"
+            - name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+              value: http://jaeger.rollout-operator-development.svc.cluster.local.:4318/v1/traces
```
colega, line 45: Note for reviewers: jaeger also supports otlp, and I didn't replace it with something else because having something that has a frontend to query comes in handy in this dev env.

PR #188, opened 2024-12-13
Make Kubernetes client timeout configurable

56quarters:
Make the timeout added in #186 configurable and default to 5 minutes.

PR #240, opened 2025-07-01
Always configure HTTP client with a timeout

narqo:
This is a fixup for #186 #188

We've recently had a reoccurring problem, where it took 30m for ingesters to respond to the `DELETE /prepare-instance-ring-downscale`.

```
duration=36m54.240982501s msg="error sending HTTP DELETE request to endpoint" err="Delete \"http://ingester-zone-c-98.ingester-zone-c.mimir-prod-24.svc.cluster.local/./ingester/prepare-instance-ring-downscale\": read tcp 10.3.188.28:37044->10.3.53.215:80: read: connection timed out"
```

Waiting for such a long time delays the reconciliation.

It's turned out that the "default" timeout that we've introduced in #186 is only applied, when the operator has a config, which we currently don't.

This PR updates the code making sure the configured timeout is always set in the HTTP client config.

### DIFF
--- a/cmd/rollout-operator/main.go
+++ b/cmd/rollout-operator/main.go
@@ -165,10 +165,13 @@ func main() {
 	check(srv.Start())
 
 	// Build the Kubernetes client config.
-	kubeConfig, err := buildKubeConfig(cfg.kubeAPIURL, cfg.kubeConfigFile, cfg.kubeClientTimeout)
+	kubeConfig, err := buildKubeConfig(cfg.kubeAPIURL, cfg.kubeConfigFile)
 	check(errors.Wrap(err, "failed to build Kubernetes client config"))
 	instrumentation.InstrumentKubernetesAPIClient(kubeConfig, reg)
 
+	if kubeConfig.Timeout == 0 {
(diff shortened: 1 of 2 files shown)


cmd/rollout-operator/main.go
```diff
@@ -165,10 +165,13 @@ func main() {
 	check(srv.Start())
 
 	// Build the Kubernetes client config.
-	kubeConfig, err := buildKubeConfig(cfg.kubeAPIURL, cfg.kubeConfigFile, cfg.kubeClientTimeout)
+	kubeConfig, err := buildKubeConfig(cfg.kubeAPIURL, cfg.kubeConfigFile)
 	check(errors.Wrap(err, "failed to build Kubernetes client config"))
 	instrumentation.InstrumentKubernetesAPIClient(kubeConfig, reg)
 
+	if kubeConfig.Timeout == 0 {
+		kubeConfig.Timeout = cfg.kubeClientTimeout
```
56quarters, line 173: Can you update the docs for this option to indicate it applies to all HTTP requests, not just Kubernetes API requests?

Issue #251, opened 2025-07-29
Prepare Downscale endpoint uses statefulset name instead of service name


gmanghna:
createPrepareDownscaleEndpoints method expects a service name but ```sts.GetName()```  is passed. Hence, constructing an invalid FQDN to the pod to downscale with delay.

 https://github.com/grafana/rollout-operator/blob/main/pkg/controller/delay.go#L64

PR #252, opened 2025-07-29
fix: Use service name for downscale endpoint

gmanghna:
Addresses https://github.com/grafana/rollout-operator/issues/251
As pointed out in the issue, similar issue is addressed for different location in code earlier.

### DIFF
--- a/pkg/controller/delay.go
+++ b/pkg/controller/delay.go
@@ -31,7 +31,7 @@ func cancelDelayedDownscaleIfConfigured(ctx context.Context, logger log.Logger, 
 		return
 	}
 
-	endpoints := createPrepareDownscaleEndpoints(sts.Namespace, sts.GetName(), 0, int(replicas), prepareURL)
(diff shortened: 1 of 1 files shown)


pkg/controller/delay.go
```diff
@@ -31,7 +31,7 @@ func cancelDelayedDownscaleIfConfigured(ctx context.Context, logger log.Logger,
 		return
 	}
 
-	endpoints := createPrepareDownscaleEndpoints(sts.Namespace, sts.GetName(), 0, int(replicas), prepareURL)
+	endpoints := createPrepareDownscaleEndpoints(sts.Namespace, sts.Spec.ServiceName, 0, int(replicas), prepareURL)
```
andyasp, line 34: Both the StatefulSet name and the service name are required to construct the FQDN. It doesn't work to swap one for the other since they can be different.

andyasp (changes requested):
Thank you for attempting to fix this. I think I'm going to try to refactor the code in another PR to use a common function instead since a very similar bug was fixed before and was missed here.

PR #254, opened 2025-07-31
fix: reject scale down request if there is already an in-process request in the same rollout group

sandeepsukhani:
If multiple simultaneous scale-down requests are made to prepare the downscale handler for the same rollout group, the code might inadvertently allow all the requests, even if `grafana.com/min-time-between-zones-downscale` is set to a non-zero value.
The issue is that there is no synchronisation to prevent that from happening.

In this PR, I am fixing the issue by tracking ongoing prepare downscale requests by rollout group and rejecting any new downscale requests for the same rollout group until the previous request is finished serving.

### DIFF
--- a/pkg/admission/zone_tracker_test.go
+++ b/pkg/admission/zone_tracker_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -317,6 +318,7 @@ func TestZoneTrackerConcurrentDownscale(t *testing.T) {
 	path := "/prepare-downscale"
 	rolloutGroupIngester := "ingester"
 	ingesterZoneA := "ingester-zone-a"
+	ingesterZoneB := "ingester-zone-b"
 	rolloutGroupIndexGateway := "index-gateway"
 	indexGatewayZoneA := "index-gateway-zone-a"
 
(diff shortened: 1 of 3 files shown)


JordanRushing (approved):
LGTM, let's test further before merging

pkg/admission/zone_tracker_test.go
```diff
@@ -286,3 +294,125 @@ func TestLastDownscaledNonExistentZone(t *testing.T) {
...
+			},
+		}
+	}
+
+	api := fake.NewSimpleClientset()
+
+	zt := newZoneTracker(api, namespace, "testconfigmap")
+
+	// mimic in progress downscaling request of rollout group ingester
+	zt.rolloutGroupDownscalingInProgress.Store(rolloutGroupIngester, ingesterZoneA)
```
charleskorn: Rather than reaching into the internals like this, what if we changed the test to something like:

1. configure StatefulSet to call fake prepare downscale handler
2. initiate downscale, rollout-operator calls fake handler and marks downscale as in progress
3. fake handler does not return until signalled later
4. initiate a second downscale: rollout-operator should reject the call immediately given the other is in progress
5. release fake handler to allow first request to complete

sandeepsukhani: Thanks Charles, for the review! I have made the suggested change.

pkg/admission/zone_tracker_test.go
```diff
@@ -378,41 +380,63 @@ func TestZoneTrackerConcurrentDownscale(t *testing.T) {
...
 	_, zoneUpdated := zt.zones[indexGatewayZoneA]
 	require.True(t, zoneUpdated)
 
-	// clearing the in progress downscaling request of rollout group ingester should let new requests to go through
-	zt.rolloutGroupDownscalingInProgress.Delete(rolloutGroupIngester)
-	ar = buildAdmissionRequest(rolloutGroupIngester, ingesterZoneA)
+	// finishing the in progress downscaling request for rollout group ingester should let new requests to go through
+	close(ingesterZoneAPrepDownscaleDone)
+	ar = buildAdmissionRequest(rolloutGroupIngester, ingesterZoneB)
 	admissionResponse = zt.prepareDownscale(context.Background(), logger, ar, api, f)
```
charleskorn: This will be racy - this request could arrive while the zone A request is still finishing. Should we wait for the zone A admission request to complete before sending the next request?

sandeepsukhani: Yeah, sorry, I missed that. I have fixed it. Thank you!

pkg/admission/zone_tracker_test.go
```diff
@@ -286,3 +295,150 @@ func TestLastDownscaledNonExistentZone(t *testing.T) {
...
+	_, zoneUpdated := zt.zones[indexGatewayZoneA]
+	require.True(t, zoneUpdated)
+
+	// finishing the in progress downscaling request for rollout group ingester should let new requests to go through
+	close(ingesterZoneAPrepDownscaleDone)
+
+	require.Eventually(t, func() bool {
+		ar = buildAdmissionRequest(rolloutGroupIngester, ingesterZoneB)
+		return zt.prepareDownscale(context.Background(), logger, ar, api, f).Allowed
+	}, 5*time.Second, time.Millisecond)
```
charleskorn: Running this every millisecond might be a little too aggressive - how about we run it a little less frequently?

```suggestion
	}, 5*time.Second, 50*time.Millisecond)
```

charleskorn (approved):
LGTM modulo suggestion below

pkg/admission/zone_tracker_test.go
sandeepsukhani: makes sense. Thanks.

PR #258, opened 2025-08-05
fix: construct delay downscale pod FQDN using both service and sts name

andyasp:
Fixes https://github.com/grafana/rollout-operator/issues/251

`createPrepareDownscaleEndpoints` assumed the StatefulSet name was equal to the serviceName. That works in Mimir's Jsonnet deployment, but does not for Mimir's Helm chart. This is similar to the fix in #221.

I pulled the FQDN construction into a common method to hopefully reduce the chance of this drifting again in the future. A consequence of doing that was the prepare downscale path now specifies a trailing `.` as well when it did not before. That is not a required change for this particular fix, but it seemed right to do anyway.


56quarters (approved):
LGTM. I imagine we're going to need to make the cluster domain configurable at some point but it doesn't need to be today.

PR #262, opened 2025-08-12
Use informer to monitor for new webhooks

tcp13equals2:
Modify the rollout-operator to use an informer to monitor for new webhook configurations.

Webhook configurations need to be patched with a self signed CA, and previously this was only checked and applied on rollout-operator startup.

This change allows for new webhooks to be created once the rollout-operator is running and have the CA patch applied.

There is a matching jsonnet update in https://github.com/grafana/mimir/pull/12360

There is a matching r355.libsonnet to enable the additional watch permission in https://github.com/grafana/deployment_tools/pull/321550

### DIFF
--- a/cmd/rollout-operator/main.go
+++ b/cmd/rollout-operator/main.go
@@ -206,10 +206,10 @@ func main() {
 	dynamicClient, err := dynamic.NewForConfigAndClient(kubeConfig, httpClient)
 	check(errors.Wrap(err, "failed to init dynamicClient"))
 
-	vwo := tlscert.NewWebhookObserver(kubeClient, cfg.kubeNamespace, logger)
+	webhookObserver := tlscert.NewWebhookObserver(kubeClient, cfg.kubeNamespace, logger)
 
 	// Start TLS server if enabled.
-	maybeStartTLSServer(cfg, httpRT, logger, kubeClient, restart, metrics, vwo)
+	maybeStartTLSServer(cfg, httpRT, logger, kubeClient, restart, metrics, webhookObserver)
 
 	// Init the controller.
 	c := controller.NewRolloutController(kubeClient, restMapper, scaleClient, dynamicClient, cfg.kubeNamespace, httpClient, cfg.reconcileInterval, reg, logger)
@@ -219,7 +219,7 @@ func main() {
 	go func() {
 		waitForSignalOrRestart(logger, restart)
 		c.Stop()
-		vwo.Stop()
+		webhookObserver.Stop()
 	}()
 
 	// The operator is ready once the controller successfully initialised.
@@ -273,10 +273,10 @@ func maybeStartTLSServer(cfg config, rt http.RoundTripper, logger log.Logger, ku
 
 		webHookListener := &tlscert.WebhookConfigurationListener{
 			OnValidatingWebhookConfiguration: func(webhook *admissionregistrationv1.ValidatingWebhookConfiguration) error {
-				return tlscert.PatchCABundleOnValidatingWebhooks(logger, kubeClient, cfg.kubeNamespace, cert.CA, webhook)
+				return tlscert.PatchCABundleOnValidatingWebhook(logger, kubeClient, cfg.kubeNamespace, cert.CA, webhook)
 			},
 			OnMutatingWebhookConfiguration: func(webhook *admissionregistrationv1.MutatingWebhookConfiguration) error {
-				return tlscert.PatchCABundleOnMutatingWebhooks(logger, kubeClient, cfg.kubeNamespace, cert.CA, webhook)
+				return tlscert.PatchCABundleOnMutatingWebhook(logger, kubeClient, cfg.kubeNamespace, cert.CA, webhook)
 			},
 		}
 
--- a/cmd/rollout-operator/main.go
+++ b/cmd/rollout-operator/main.go
@@ -219,6 +219,7 @@ func main() {
 	go func() {
 		waitForSignalOrRestart(logger, restart)
 		c.Stop()
+		vwo.Stop()
 	}()
 
 	// The operator is ready once the controller successfully initialised.
--- a/cmd/rollout-operator/main.go
+++ b/cmd/rollout-operator/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/prometheus/common/model"
 	"go.uber.org/atomic"
 	v1 "k8s.io/api/admission/v1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // Required to get the GCP auth provider working.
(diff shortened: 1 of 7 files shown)


cmd/rollout-operator/main.go
```diff
@@ -205,8 +206,10 @@ func main() {
 	dynamicClient, err := dynamic.NewForConfigAndClient(kubeConfig, httpClient)
 	check(errors.Wrap(err, "failed to init dynamicClient"))
 
+	vwo := tlscert.NewWebhookObserver(kubeClient, cfg.kubeNamespace, logger)
```
charleskorn: [nit] Perhaps `webhookObserver` would be a clearer variable name?

pkg/tlscert/webhook_observer.go
```diff
@@ -0,0 +1,95 @@
...
+type WebhookConfigurationListener struct {
+	OnValidatingWebhookConfiguration func(webhook *admissionregistrationv1.ValidatingWebhookConfiguration) error
+	OnMutatingWebhookConfiguration   func(webhook *admissionregistrationv1.MutatingWebhookConfiguration) error
+}
+
+// A WebhookObserver uses informers to monitor for ValidatingWebhookConfiguration and MutatingWebhookConfigurations.
+// When observed, a call back function is invoked with the given configuration.
+// The call back function is set as part of the Init() function.
+type WebhookObserver struct {
+	namespace                  string
```
charleskorn: Is this field needed? Looks like it's not used anywhere.

pkg/tlscert/webhook_observer.go
```diff
@@ -0,0 +1,95 @@
...
+					err = onEvent.OnValidatingWebhookConfiguration(validatingWebHook)
+				} else {
+					mutatingWebHook, isObj := obj.(*admissionregistrationv1.MutatingWebhookConfiguration)
+					if isObj {
+						err = onEvent.OnMutatingWebhookConfiguration(mutatingWebHook)
+					}
+				}
+				if err != nil {
+					level.Error(c.log).Log("msg", "Unable to call webhook configuration listener", "err", err)
+				}
```
charleskorn: This would be clearer as a type switch: https://go.dev/tour/methods/16

pkg/tlscert/webhook_observer.go
```diff
@@ -0,0 +1,95 @@
...
+			return errors.Wrap(err, "failed to add webhook listener")
+		}
+	}
+
+	go c.factory.Start(c.stopCh)
+
+	// Wait until all informer caches have been synced.
+	level.Info(c.log).Log("msg", "waiting for webhook informer caches to sync")
+	if ok := cache.WaitForCacheSync(c.stopCh, c.informerValidatingWebhooks.HasSynced, c.informerMutatingWebhooks.HasSynced); !ok {
+		return errors.New("validating webhook informer caches failed to sync")
```
charleskorn: Is there any more details of what went wrong that we can include here?

pkg/tlscert/webhook_observer_test.go
```diff
@@ -0,0 +1,123 @@
...
+func mutatingWebhookConfiguration(name string) *admissionregistrationv1.MutatingWebhookConfiguration {
+	return &admissionregistrationv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+}
+
+// await - sleep for a short period and invoke the given task. return if the task returns true. repeat for a fixed number of iterations
+func await(t *testing.T, task func() bool, message string) {
```
charleskorn: Can we use `require.Eventually` for this, instead of implementing our own?

pkg/tlscert/webhookconfig.go
```diff
@@ -18,100 +19,81 @@ import (
 // Webhook configurations should have the following labels:
 // "grafana.com/inject-rollout-operator-ca": "true",
 // "grafana.com/namespace":                  <specified namespace>,
-func PatchCABundleOnValidatingWebhooks(ctx context.Context, logger log.Logger, kubeClient kubernetes.Interface, namespace string, caPEM []byte) error {
+func PatchCABundleOnValidatingWebhooks(logger log.Logger, kubeClient kubernetes.Interface, namespace string, caPEM []byte, wh *v1.ValidatingWebhookConfiguration) error {
```
charleskorn: [nit] This method now only handles a single webhook:
```suggestion
func PatchCABundleOnValidatingWebhook(logger log.Logger, kubeClient kubernetes.Interface, namespace string, caPEM []byte, wh *v1.ValidatingWebhookConfiguration) error {
```

(same applies for `PatchCABundleOnMutatingWebhooks`)

pkg/tlscert/webhook_observer_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	_, err := client.AdmissionregistrationV1().ValidatingWebhookConfigurations().Create(context.Background(), validatingWebhookConfiguration("validating-webhook"), metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// wait for the informer to be aware of this create
+	task := func() bool {
+		return len(observedValidatingWebhooks) == 1
+	}
+	await(t, task, "ValidatingWebhookConfiguration should have 1 ValidatingWebhookConfiguration")
+	require.Equal(t, 1, len(observedValidatingWebhooks))
```
charleskorn: This assertion seems unnecessary given it's the condition for `await` to succeed (same applies for other `await` call below)

pkg/tlscert/webhook_observer_test.go
```diff
@@ -0,0 +1,123 @@
...
+	require.Equal(t, "validating-webhook", observedValidatingWebhooks[0].Name)
+	require.Equal(t, 0, len(observedMutatingWebhooks))
```
charleskorn: [nit]

```suggestion
	require.Empty(t, observedMutatingWebhooks)
```

pkg/tlscert/webhook_observer_test.go
```diff
@@ -0,0 +1,123 @@
...
+		OnMutatingWebhookConfiguration: func(webhook *admissionregistrationv1.MutatingWebhookConfiguration) error {
+			observedMutatingWebhooks = append(observedMutatingWebhooks, webhook)
+			return nil
+		},
+	}
+	require.NoError(t, observer.Init(listener))
+	defer observer.Stop()
+
+	require.Equal(t, 0, len(observedValidatingWebhooks))
+	require.Equal(t, 0, len(observedMutatingWebhooks))
```
charleskorn: [nit]

```suggestion
	require.Empty(t, observedValidatingWebhooks)
	require.Empty(t, observedMutatingWebhooks)
```

pkg/tlscert/webhook_observer.go
tcp13equals2: Is there something specific you have in mind? 

This logging matches the similar WaitForCacheSync in the main controller.go

pkg/tlscert/webhook_observer_test.go
tcp13equals2: Nice!

pkg/tlscert/webhook_observer.go
charleskorn: Nothing specific, I was just hoping there was some way to include more information here.

charleskorn (approved):
LGTM, thanks for working on this!

pkg/tlscert/webhook_observer.go
tcp13equals2: I think all I can really add is extra info on which cache has not synced. So have included that in the error message

andyasp:
Thanks @tcp13equals2. Could you open a follow-up PR to add a [CHANGELOG](https://github.com/grafana/rollout-operator/blob/main/CHANGELOG.md) entry for this change? Other teams besides Mimir use rollout-operator so it would help with communicating the `watch` change.

Issue #261, opened 2025-08-11
Support a non-default cluster domain


andyasp:
As mentioned in https://github.com/grafana/rollout-operator/pull/258#pullrequestreview-3107134627, when constructing pod endpoints the cluster domain is hardcoded to the default `cluster.local`. It could be nice to support a non-default value.

[Code reference](https://github.com/grafana/rollout-operator/blob/48b0e5c0f0acf4d92f0649fd0077430c5bb04c9c/pkg/util/util.go#L107C67-L120C2):
```go
func StatefulSetPodFQDN(namespace, statefulSetName string, ordinal int, serviceName string) string {
	// The DNS entry for a pod of a stateful set is
	// $(statefulset name)-(ordinal).$(service name).$(namespace).svc.cluster.local
	// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
	//
	// Ending that with a trailing "." to signify an absolute domain name
	// https://datatracker.ietf.org/doc/html/rfc1034#section-3.1
	return fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local.",
		statefulSetName,
		ordinal,
		serviceName,
		namespace,
	)
}
```

Kubernetes documentation: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id

PR #266, opened 2025-08-14
Support a non-default cluster domain

andyasp:
Fixes https://github.com/grafana/rollout-operator/issues/261

When building the endpoints for StatefulSet pods, the cluster domain was always assumed to be `cluster.local`. This change makes the cluster domain configurable (with a default of `cluster.local`) and does the required plumbing to get the configured value to where it used.

Opening in draft because I'd rather wait on this until after the next release is cut.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -104,17 +104,18 @@ func Now() *metav1.Time {
 	return &ts
 }
 
-func StatefulSetPodFQDN(namespace, statefulSetName string, ordinal int, serviceName string) string {
+func StatefulSetPodFQDN(namespace, statefulSetName string, ordinal int, serviceName, clusterDomain string) string {
 	// The DNS entry for a pod of a stateful set is
-	// $(statefulset name)-(ordinal).$(service name).$(namespace).svc.cluster.local
+	// $(statefulset name)-(ordinal).$(service name).$(namespace).svc.$(cluster domain)
 	// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
 	//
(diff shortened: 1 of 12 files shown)


pkg/util/util.go
```diff
@@ -104,17 +104,18 @@ func Now() *metav1.Time {
...
+func StatefulSetPodFQDN(namespace, statefulSetName string, ordinal int, serviceName, clusterDomain string) string {
 	// The DNS entry for a pod of a stateful set is
-	// $(statefulset name)-(ordinal).$(service name).$(namespace).svc.cluster.local
+	// $(statefulset name)-(ordinal).$(service name).$(namespace).svc.$(cluster domain)
 	// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id
 	//
 	// Ending that with a trailing "." to signify an absolute domain name
 	// https://datatracker.ietf.org/doc/html/rfc1034#section-3.1
-	return fmt.Sprintf("%v-%v.%v.%v.svc.cluster.local.",
+	return fmt.Sprintf("%v-%v.%v.%v.svc.%s.",
```
56quarters: I think it's reasonable to support both domains that end with the root and those that don't: `example.com.` vs `example.com`. Can we adjust the configuration defaults and parsing to make that possible (or the logic here)?

andyasp: That makes sense, thanks. I checked what we do in the Mimir Helm chart and the default value is
```
clusterDomain: cluster.local.
```
So I moved the default from `cluster.local` to `cluster.local.` and modified the function to not assume the ending `.`.

PR #273, opened 2025-08-25
Fix issue of observed webhook object changing prior to CA patch

tcp13equals2:
Issue observed where the patch failed because the object had changed from when the webhook was observed and the listener applied the patch. 

Added an ability to retry if this error is detected.

Also changed the observer to also action update events so if the initial create/insert fails an update can still attempt the CA patch.

### DIFF
--- a/pkg/tlscert/webhook_observer.go
+++ b/pkg/tlscert/webhook_observer.go
@@ -2,6 +2,7 @@ package tlscert
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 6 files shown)


pkg/tlscert/webhook_observer.go
```diff
@@ -45,31 +48,50 @@ func NewWebhookObserver(kubeClient kubernetes.Interface, namespace string, logge
...
 
 	return c
 }
 
+func (c *WebhookObserver) onWebHookConfigurationObserved(obj interface{}) {
+
+	if c.onEvent == nil {
+		level.Error(c.log).Log("msg", "webhook configuration listener has not been initialized")
+		return
+	}
```
charleskorn: I don't think this is necessary - if `onEvent` is `nil`, something is wrong and panicking is fine.

charleskorn (commented):
LGTM. Could you please update the changelog entry for https://github.com/grafana/rollout-operator/pull/262 to reference this PR as well?

charleskorn (approved):

PR #278, opened 2025-08-27
Only automatically build images for pushed tags

andyasp:
Image builds in CI are taking 50m+ each on every commit to `main`. This happens even for changes that don't touch the code like github action updates. It would be nice to make the image builds faster, but for now I think restricting the image builds to tag pushes makes sense. If someone needs an image between releases they can still build one manually.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -88,7 +88,7 @@ jobs:
       - integration
       - integration-boringcrypto
       - lint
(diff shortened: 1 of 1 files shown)


charleskorn (approved):
Not blocking: perhaps we could skip boringcrypto image builds for `main`?

andyasp:
That sounds good to me. What we could try if we had that too is having two Dockerfiles, one for the boringcrypto build and one for the normal build, but the normal build could cross-compile then.

andyasp:
I'll merge this for now.

PR #282, opened 2025-09-01
Move rollout-operator libsonnet from mimir

tcp13equals2:
This PR relates to https://github.com/grafana/mimir-squad/issues/3179.

The jsonnet related to the rollout-operator has been copied from the mimir repo and made more re-usable for other projects which also use the rollout-operator.

Once this PR is merged the mimir jsonnet will be updated accordingly to use this new jsonnet.

Similar style jsonnet tests have been created, with a focus on the validation of the jsonnet generated for the rollout-operator.

The rollout-operator integration tests have been modified to use dynamically generated manifest values using jsonnet configurations.


.github/workflows/ci.yaml
```diff
@@ -80,6 +80,35 @@ jobs:
...
+  lint-jsonnet:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+        with:
+          persist-credentials: false
+      - name: Install jsonnet
+        run: | 
+          sudo apt-get update
+          sudo apt-get install -y jsonnet
```
charleskorn: Is there any way we can pin this to a particular version, to avoid unexpected upgrades suddenly breaking things?

.github/workflows/ci.yaml
```diff
@@ -80,6 +80,35 @@ jobs:
...
+            | tar zx -C /tmp
+          sudo mv /tmp/conftest /usr/local/bin/
+      - name: Install tk
+        run: |
+          sudo curl -Lo /usr/local/bin/tk https://github.com/grafana/tanka/releases/latest/download/tk-linux-amd64
+          sudo chmod a+x /usr/local/bin/tk
+      - name: Install jb
+        run: |
+          sudo curl -Lo /usr/local/bin/jb https://github.com/jsonnet-bundler/jsonnet-bundler/releases/latest/download/jb-linux-amd64
+          sudo chmod a+x /usr/local/bin/jb
```
charleskorn: Same as above - would be good to pin these versions

.github/workflows/ci.yaml
```diff
@@ -80,6 +80,35 @@ jobs:
...
+      - name: Check Jsonnet Manifests
+        run: make BUILD_IN_CONTAINER=false check-jsonnet-manifests
+      - name: Check Jsonnet Tests
+        run: make BUILD_IN_CONTAINER=false check-jsonnet-tests
```
charleskorn: Is `BUILD_IN_CONTAINER` needed here?

```suggestion
      - name: Check Jsonnet Manifests
        run: make check-jsonnet-manifests
      - name: Check Jsonnet Tests
        run: make check-jsonnet-tests
```

Makefile
```diff
@@ -86,3 +101,41 @@ fix-lint: ## Automatically fix linting issues where possible
...
+
+.PHONY: check-jsonnet-manifests
+check-jsonnet-manifests: ## Check the rollout-operator manifests.
+check-jsonnet-manifests: format-jsonnet-manifests
+	@echo "Checking diff:"
+	@./tools/find-diff-or-untracked.sh $(JSONNET_MANIFESTS_PATHS) || (echo "Please format jsonnet manifests by running 'make format-jsonnet-manifests'" && false)
+
+.PHONY: format-jsonnet-manifests
+format-jsonnet-manifests: ## Format the rollout-operator manifests.
+	@find $(JSONNET_MANIFESTS_PATHS) -type f -name '*.libsonnet' -print -o -name '*.jsonnet' -print | xargs $(JSONNET_FMT) -i
```
charleskorn:
```suggestion
.PHONY: format-jsonnet
format-jsonnet: ## Format the rollout-operator manifests.
	@find $(JSONNET_MANIFESTS_PATHS) -type f -name '*.libsonnet' -print -o -name '*.jsonnet' -print | xargs $(JSONNET_FMT) -i
```

Makefile
```diff
@@ -86,3 +101,41 @@ fix-lint: ## Automatically fix linting issues where possible
...
+
+.PHONY: build-jsonnet-tests
+build-jsonnet-tests: ## Build the rollout-operator tests.
+	@./operations/rollout-operator-tests/build.sh
+
+.PHONY: check-jsonnet-tests
+check-jsonnet-tests: ## Check the jsonnet tests output.
+check-jsonnet-tests: build-jsonnet-tests jsonnet-conftest-test
+	@./tools/find-diff-or-untracked.sh operations/mimir-tests || (echo "Please rebuild jsonnet tests output 'make build-jsonnet-tests'" && false)
```
charleskorn:
```suggestion
	@./tools/find-diff-or-untracked.sh operations/rollout-operator-tests || (echo "Please rebuild jsonnet tests output 'make build-jsonnet-tests'" && false)
```

Makefile
```diff
@@ -86,3 +101,41 @@ fix-lint: ## Automatically fix linting issues where possible
...
+.PHONY: check-makefiles
+check-makefiles: ## Check the makefiles format.
+check-makefiles: format-makefiles
+	echo $(MAKE_FILES)
+	@git diff --exit-code -- $(MAKE_FILES) || (echo "Please format Makefiles by running 'make format-makefiles'" && false)
+
+.PHONY: format-makefiles
+format-makefiles: ## Format all Makefiles.
+format-makefiles: $(MAKE_FILES)
+	$(SED) -i -e 's/^\(\t*\)  /\1\t/g' -e 's/^\(\t*\) /\1/' -- $?
```
charleskorn: This seems like an unrelated change, I would move this to a separate PR (and also add it to the CI workflow)

operations/rollout-operator/crds/replica-templates.yaml
charleskorn, line 1: Should we remove the copies of this file and the ZPDB CRD in the `development` directory and use these copies instead? Otherwise I worry they'll get out of sync.

operations/rollout-operator/images.libsonnet
charleskorn, line 1: We should add a note to the release instructions on updating this file before tagging the release in Git.

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,380 @@
...
+
+    // Configure the rollout operator to accept webhook requests made as part of scaling
+    // statefulsets up or down. This allows the rollout operator to ensure that stateful
+    // components (ingesters, store-gateways) are scaled up or down safely.
+    enable_rollout_operator_webhook: false,
+
+    // Specifically enable zone aware pod distruption budget support, including the rollout-operator webhooks.
+    // Note that zpdb support is enabled by default if webhooks are enabled.
+    // See disable_rollout_operator_eviction_webhook for additional configuration options.
+    enable_zpdb: false,
```
charleskorn: Should these default to true?

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,380 @@
...
+
+    // Ignore these labels used for controlling webhook behavior when creating services.
+    service_ignored_labels+:: ['grafana.com/no-downscale', 'grafana.com/prepare-downscale'],
+  },
+
+  // implicitly enable webhooks if the zpdb or replica template support has been requested
+  local enableWebhooks =
+    $._config.enable_zpdb ||
+    $._config.rollout_operator_replica_template_access_enabled ||
+    $._config.enable_rollout_operator_webhook,
```
charleskorn: Should we use an assertion to just fail if ZPDBs are enabled but webhooks aren't?

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,380 @@
...
+
+  // implicitly enable rollout-operator if the webhooks have been requested
+  local enableRolloutOperator = enableWebhooks || $._config.rollout_operator_enabled,
```
charleskorn: Similar to above - if webhooks are enabled but the rollout-operator as a whole isn't, then I'd either fail with an assertion message or ignore the webhooks and create nothing.

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,380 @@
...
+    enable_zpdb: false,
+
+    // Disable the deployment of the pod eviction webhook.
+    // This will only have an effect if webhooks are enabled.
+    // Note that the eviction webhook is safe to be deployed even if there are no zone pdb configurations.
+    // Pod evictions not handled by a zpdb will be allowed and contine to be handled by any other pdb configurations.
+    disable_rollout_operator_eviction_webhook: false,
+
+    // Include the custom resource definiton for the zpdb.
+    zpdb_custom_resource_definition_enabled: true,
```
charleskorn: Do we need this as well as `enable_zpdb`? Seems unnecessary.

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,380 @@
...
+    // Specifically enable zone aware pod distruption budget support, including the rollout-operator webhooks.
+    // Note that zpdb support is enabled by default if webhooks are enabled.
+    // See disable_rollout_operator_eviction_webhook for additional configuration options.
+    enable_zpdb: false,
+
+    // Disable the deployment of the pod eviction webhook.
+    // This will only have an effect if webhooks are enabled.
+    // Note that the eviction webhook is safe to be deployed even if there are no zone pdb configurations.
+    // Pod evictions not handled by a zpdb will be allowed and contine to be handled by any other pdb configurations.
+    disable_rollout_operator_eviction_webhook: false,
```
charleskorn: Under what scenario do we need this if `ignore_rollout_operator_zpdb_eviction_webhook_failures` is set to true? This one seems unnecessary.

operations/policies/policies.rego
```diff
@@ -0,0 +1,54 @@
...
+	count(keys) == 0
+}
+
+deny contains msg if {
+	obj := input[i].contents
+	msg := sprintf("ServiceAccount does not have expected name %v expected=%v got %v", [object_display_name[i], "rollout-operator", obj.metadata.name])
+
+	obj.kind == "ServiceAccount"
+	obj.metadata.name != "rollout-operator"
+}
```
charleskorn: Do we need this policy? What is it protecting against?

tcp13equals2: That service account name is referenced in the deployment spec.template.spec.serviceAccountName. 

So this is checking that these names match and there by avoid a rollout-operator pod failing to start because the service account name could not be found.

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: It's there as an option for users who may be using the rollout-operator webhooks but not any zpdb configurations, and they want to avoid all pod eviction requests hair-pinning through the pod-eviction webhook. 

Some users might not like putting all evictions through a webhook which is not doing anything - even if it's declared safe.

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: In the existing mimir codebase it is set as the following ... Do you think it would be safe to change this existing default behavior?  

`
enable_rollout_operator_webhook: false,

rollout_operator_enabled: $._config.multi_zone_ingester_enabled ||
                              $._config.multi_zone_store_gateway_enabled ||
                              $._config.cortex_compactor_concurrent_rollout_enabled ||
                              $._config.ingest_storage_ingester_autoscaling_enabled ||
                              $._config.enable_rollout_operator_webhook,
`

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: There is precedent in the existing mimir code base for the implicit enabling based off other variables ... 

Lets discuss in our call today.

`
rollout_operator_enabled: $._config.multi_zone_ingester_enabled ||
$._config.multi_zone_store_gateway_enabled ||
$._config.cortex_compactor_concurrent_rollout_enabled ||
$._config.ingest_storage_ingester_autoscaling_enabled ||
$._config.enable_rollout_operator_webhook,
`

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: We need to keep zpdb_custom_resource_definition_enabled, but I have removed the enable_zpdb

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: I've removed this.

.github/workflows/ci.yaml
```diff
@@ -80,6 +80,39 @@ jobs:
...
+          sudo mv jsonnet jsonnetfmt /usr/local/bin
+      - name: Install conftest
+        run: |
+          CONTEST_VERSION="0.62.0"
+          curl -sL "https://github.com/open-policy-agent/conftest/releases/download/v${CONTEST_VERSION}/conftest_${CONTEST_VERSION}_Linux_x86_64.tar.gz" \
+            | sudo tar -xz -C /usr/local/bin
+      - name: Install tk
+        run: |
+          TANKER_VERSION="0.34.1"
+          sudo curl -sLo /usr/local/bin/tk https://github.com/grafana/tanka/releases/download/v${TANKER_VERSION}/tk-linux-amd64
```
charleskorn:
```suggestion
          TANKA_VERSION="0.34.1"
          sudo curl -sLo /usr/local/bin/tk https://github.com/grafana/tanka/releases/download/v${TANKA_VERSION}/tk-linux-amd64
```

operations/rollout-operator/rollout-operator.libsonnet
charleskorn: I would have them default to enabled here, and then have Mimir override the defaults to suit Mimir.

operations/policies/policies.rego
charleskorn: OK, in that case, what if we had a policy that checks that the service account name matches the service account name given in the deployment? Then the link between them is clearer.

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: Per our discussion I have enabled the rollout-operator and its webhooks by default.

.github/workflows/ci.yaml
```diff
@@ -80,6 +80,39 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+        with:
+          persist-credentials: false
+      - name: Install jsonnet
+        run: |
+          JSONNET_VERSION=0.21.0
+          curl -sL "https://github.com/google/jsonnet/releases/download/v${JSONNET_VERSION}/jsonnet-v${JSONNET_VERSION}.tar.gz" | tar -xz
+          cd jsonnet-v${JSONNET_VERSION}
+          make > /dev/null
```
charleskorn: Any reason to hide the output here?

charleskorn: We could also use `go-jsonnet` and its precompiled binaries here to save some time.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main / unreleased
 
+* [ENHANCEMENT] Move rollout-operator libsonnet into this repository. Enable rollout-operator and its webhooks by default. #282
+* [CHANGE] Rename flag `enable_rollout_operator_webhook` to `rollout_operator_webhooks_enabled`. #282
+* [CHANGE] Add flag compatibility checks to prevent invalid flag combinations. #282
```
charleskorn: Given the Jsonnet bundle wasn't in this repository before, I would describe this as an addition (rather than a move), and I don't think we need to list these as changes here.

RELEASE.md
```diff
@@ -16,3 +16,5 @@
 4. Update the Helm Chart:
    - Repository https://github.com/grafana/helm-charts/tree/main/charts/rollout-operator
    - [Example PR](https://github.com/grafana/helm-charts/pull/3177/files)
+
+5. Update [images.libsonnet] (operations/rollout-operator/images.libsonnet)
```
charleskorn: Should this go between the current steps 1 and 2 so that it's included in the tagged version?

operations/policies/jsonnet.yaml
```diff
@@ -0,0 +1,2 @@
+# We currently don't support this for Jsonnet.
+disable_restricted_security_context_checks: true
```
charleskorn: This doesn't seem to be used anywhere - should it be removed?

operations/policies/policies_deployment.rego
```diff
@@ -0,0 +1,55 @@
+package main
+
+# Tests which relate to the expected Deployment manifests
+
+is_rollout_operator_deployment_valid(obj) if {
```
charleskorn: I don't think checks like this one are valuable - they become change detector tests rather than alerting us to specific problems. Are there specific things that are important to check here?

operations/rollout-operator/crds/zone-aware-pod-disruption-budget-crd.yaml
charleskorn, line 1: Naming nit: the other file in this directory doesn't have a `-crd` prefix - should both have this prefix (or both not have this prefix)?

My 2c is that the prefix is unnecessary given the directory is called `crds`.

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -0,0 +1,371 @@
...
+      error 'ignore_rollout_operator_prepare_downscale_webhook_failures requires rollout_operator_webhooks_enabled=true'
+    else if $._config.ignore_rollout_operator_zpdb_eviction_webhook_failures && !$._config.rollout_operator_webhooks_enabled then
+      error 'ignore_rollout_operator_zpdb_eviction_webhook_failures requires rollout_operator_webhooks_enabled=true'
+    else if $._config.ignore_rollout_operator_zpdb_validation_webhook_failures && !$._config.rollout_operator_webhooks_enabled then
+      error 'ignore_rollout_operator_zpdb_validation_webhook_failures requires rollout_operator_webhooks_enabled=true'
+    else if $._config.zpdb_custom_resource_definition_enabled && !$._config.rollout_operator_webhooks_enabled then
+      error 'zpdb_custom_resource_definition_enabled requires rollout_operator_webhooks_enabled=true'
+    else if $._config.replica_template_custom_resource_definition_enabled && !$._config.rollout_operator_webhooks_enabled then
+      error 'replica_template_custom_resource_definition_enabled requires rollout_operator_webhooks_enabled=true'
```
charleskorn: Can these be rewritten as `assert` statements outside the context of `enableWebhooks`? It would help make the value of `enableWebhooks` clearer.

operations/rollout-operator/replica-template.libsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  // replicaTemplate creates new ReplicaTemplate resource.
+  // If replicas is > 0, spec.replicas field is specified in the resource, if replicas <= 0, spec.replicas field is hidden.
+  // Syntactically valid label selector is required, and may be used by HorizontalPodAutoscaler controller when ReplicaTemplate
+  // is used as scaled resource depending on metric target type.
+  // (When using targetType=AverageValue, label selector is not used for scaling computation).
+  replicaTemplate(name, replicas, label_selector):: {
+    assert $._config.rollout_operator_webhooks_enabled : 'replica template configuration requires rollout_operator_webhooks_enabled=true',
+    apiVersion: 'rollout-operator.grafana.com/v1',
```
charleskorn: Should this be derived from `$.replica_template` too, like `kind` below?

operations/rollout-operator/zone-aware-pod-disruption-budget.libsonnet
```diff
@@ -0,0 +1,40 @@
+{
+  // newZPDB creates new ZoneAwarePodDisruptionBudget resource.
+  // name - a unique name within the namespace for this resource. eg ingester-rollout
+  // rolloutGroup - the rollout-group label to match pods who are within this zpdb. eg ingester
+  // maxUnavailable - the max unavailable pod count. Can be expressed as an integer >= 0 or as a percentage 50% (0-100%)
+  // podNamePartitionRegex - use to enable partition awareness. A regex which can extract a partition name from a pod name. ie [a-z\\-]+-zone-[a-z]-([0-9]+)
+  // podNameRegexGroup - use to define the regex grouping which holds the partition name. default is 1
+  newZPDB(name, rolloutGroup, maxUnavailable, podNamePartitionRegex='', podNameRegexGroup=1):: {
+    assert $._config.rollout_operator_webhooks_enabled : 'zpdb configuration requires rollout_operator_webhooks_enabled=true',
+    apiVersion: 'rollout-operator.grafana.com/v1',
```
charleskorn: Should this be derived from `$.replica_template` too, like `kind` below?

.github/workflows/ci.yaml
tcp13equals2: Just avoiding noise in the logs. Will remove

operations/policies/jsonnet.yaml
tcp13equals2: It seems to be used by tk - when I tried taking it out the building of the jsonnet tests failed.

operations/policies/policies_deployment.rego
tcp13equals2: I tried to only keep attributes which I did not think would change and are important to the running of this deployment.

ie the rollingUpdate strategy maxSurge is important for making sure we only have the one instance running in a rollout, the CLI args we don't want to  lose, and we should not be touching the image spec.

Putting them in here re-enforces that these are important and brings specific attention if they are changed. More attention then just seeing a diff on the generated test files.

operations/policies/jsonnet.yaml
tcp13equals2: It was referenced by conftest ... but does not seem to be needed. Will remove

operations/policies/policies_deployment.rego
charleskorn: In that case, I would create very specific tests (with correspondingly specific failure messages) to emphasise the importance of those attributes. 

As it stands, if someone changes `maxSurge` (or much of the other config on the Deployment), they'll get a `Deployment does not have expected attributes` failure message, which doesn't convey the importance of this.

charleskorn: For example, I would remove all of the checks about ports, arguments and the image, I don't think these are valuable.

operations/policies/policies_replica_template_crd.rego
```diff
@@ -0,0 +1,40 @@
+package main
+
+# Tests which relate to the expected replica template custom resource definition
+
+# Assert that the replica template custom resource definition has key attributes
+validate_replica_template_crd(obj) if {
```
charleskorn: Same as above, I don't think this test is valuable.

operations/policies/policies_zpdb_crd.rego
```diff
@@ -0,0 +1,44 @@
+package main
+
+# Tests which relate to the expected zpdb custom resource definition
+
+# Assert that the zpbd custom resource definition has key attributes
+validate_zpdb_crd(obj) if {
```
charleskorn: Same as above, I don't think this test is valuable.

charleskorn (commented):
As an alternative to the Rego policies, have you considered an integration-style test that deploys the generated manifests from one of the test cases and does a basic smoke test? I think that would be more reliable and valuable than most of the Rego policies.

operations/policies/policies_webhooks.rego
```diff
@@ -0,0 +1,80 @@
...
+webhook_rule_present(obj, expected) if {
+  rule := obj.webhooks[0].rules[_]
+  rule.apiGroups == expected.apiGroups
+  rule.apiVersions == expected.apiVersions
+  rule.operations == expected.operations
+  rule.resources == expected.resources
+  rule.scope == "Namespaced"
+}
+
+validate_webhook(obj, name_prefix, path, side_effect, match_policy, rule) if {
```
charleskorn: Same as for the deployment policies, I would break this down into smaller checks that focus on the key attributes and can provide more detailed error messages.

operations/rollout-operator-integration/build.sh
charleskorn, line 1: I would move this directory to the root `integration` directory given that's where the integration tests live.

charleskorn, line 1: How long does this script take to run? Should we run it automatically when the tests run?

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -113,7 +114,11 @@
       { name: 'rollout-operator' },
       servicePort.newNamed('https', 443, 8443) +
       servicePort.withProtocol('TCP'),
-    ),
```
charleskorn: [nit] Is there a `service.mixin.metadata.withNamespace(...)` we can use?

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -357,5 +362,9 @@
     ]),
 
   rollout_operator_service_account: if !$._config.rollout_operator_enabled then null else
-    serviceAccount.new('rollout-operator'),
+    serviceAccount.new('rollout-operator') + {
+                                                  metadata+: {
```
charleskorn: Same as above - is there a `withNamespace` mixin we can use?

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
-    obj.webhooks[0].failurePolicy in {"Fail","Ignore"}
-    object.get(obj.webhooks[0], "matchPolicy", "") == match_policy
-    obj.webhooks[0].sideEffects == side_effect
+# Assert that all ValidatingWebhookConfiguration have the expected inject-rollout-operator-ca label
+deny contains msg if {
+    obj := input[_].contents
+    obj.kind == "ValidatingWebhookConfiguration"
+    object.get(obj.metadata.labels, "grafana.com/inject-rollout-operator-ca", "false") != "true"
+    msg := sprintf("ValidatingWebhookConfiguration should have label grafana.com/inject-rollout-operator-ca=true, %v", [display_name(obj)])
+}
```
charleskorn: Is this needed? Won't the integration tests fail without this now?

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
+    msg := sprintf("ValidatingWebhookConfiguration clientConfig.service.name should match grafana.com/namespace label, %v. %v != %v", [display_name(obj), obj.webhooks[0].clientConfig.service.namespace, object.get(obj.metadata.labels, "grafana.com/namespace", "")])
+}
+
+# Assert that all ValidatingWebhookConfiguration clientConfig.service.port are 443
+deny contains msg if {
+    obj := input[_].contents
+    obj.kind == "ValidatingWebhookConfiguration"
+    obj.webhooks[0].clientConfig.service.port != 443
+    msg := sprintf("ValidatingWebhookConfiguration has unexpected clientConfig.service.port, %v. Expected 443, got %v", [display_name(obj), obj.webhooks[0].clientConfig.service.port])
+}
```
charleskorn: Is this needed? Won't the integration tests fail without this now?

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
+    msg := sprintf("ValidatingWebhookConfiguration has unexpected webhook sideEffect , %v. Expected None, got %v", [display_name(obj), obj.webhooks[0].sideEffects])
+}
+
+# Assert that all MutatingWebhookConfiguration have the expected inject-rollout-operator-ca label
+deny contains msg if {
+    obj := input[_].contents
+    obj.kind == "MutatingWebhookConfiguration"
+    object.get(obj.metadata.labels, "grafana.com/inject-rollout-operator-ca", "false") != "true"
+    msg := sprintf("MutatingWebhookConfiguration should have label grafana.com/inject-rollout-operator-ca=true, %v", [display_name(obj)])
 }
```
charleskorn: Is this needed? Won't the integration tests fail without this now?

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
+    msg := sprintf("ValidatingWebhookConfiguration should have label grafana.com/inject-rollout-operator-ca=true, %v", [display_name(obj)])
+}
+
+# Assert that all ValidatingWebhookConfiguration clientConfig.service.namespace should match the grafana.com/namespace label
+deny contains msg if {
+    obj := input[_].contents
+    obj.kind == "ValidatingWebhookConfiguration"
+    obj.webhooks[0].clientConfig.service.namespace != object.get(obj.metadata.labels, "grafana.com/namespace", "")
+    msg := sprintf("ValidatingWebhookConfiguration clientConfig.service.name should match grafana.com/namespace label, %v. %v != %v", [display_name(obj), obj.webhooks[0].clientConfig.service.namespace, object.get(obj.metadata.labels, "grafana.com/namespace", "")])
+}
```
charleskorn: This largely duplicates the similar rule for `MutatingWebhookConfiguration` - can we combine them somehow?

(similar feedback applies to other rules)

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
+
+# Assert that all MutatingWebhookConfiguration clientConfig.service.port are 443
+deny contains msg if {
+    obj := input[_].contents
+    obj.kind == "MutatingWebhookConfiguration"
+    obj.webhooks[0].clientConfig.service.port != 443
+    msg := sprintf("MutatingWebhookConfiguration has unexpected clientConfig.service.port, %v. Expected 443, got %v", [display_name(obj), obj.webhooks[0].clientConfig.service.port])
+}
+
```
charleskorn: Is this needed? Won't the integration tests fail without this now?

operations/policies/policies_webhooks.rego
```diff
@@ -11,52 +11,141 @@ webhook_rule_present(obj, expected) if {
...
+# Assert that the zpdb-validation has expected rules
 deny contains msg if {
     rule := { "apiGroups": ["rollout-operator.grafana.com"], "apiVersions": ["v1"], "operations": ["CREATE", "UPDATE"], "resources": ["zoneawarepoddisruptionbudgets"] }
 	obj := input[_].contents
 	is_zpdb_validation_webhook(obj)
-    not validate_webhook(obj, "zpdb-validation", "/admission/zpdb-validation", "None", "", rule)
-    msg := sprintf("zpdb-validation ValidatingWebhookConfiguration does not have expected attributes, %v", [display_name(obj)])
+    not webhook_rule_present(obj, rule)
+    msg := sprintf("ValidatingWebhookConfiguration does not have expected rule, %v. Expected %v, got %v", [display_name(obj), rule, obj.webhooks[0].rules])
 }
```
charleskorn: Are these needed? Won't the integration tests fail without these attributes set correctly now?

operations/policies/policies_webhooks.rego
tcp13equals2: As discussed, I would prefer to keep these sort of rego tests here. They reflect important configuration and unlikely to be changed. Although the integration tests would pick up a change here, the integration tests are a lot harder and slower to debug.

operations/policies/policies_webhooks.rego
tcp13equals2: Have updated to compare it to the Service port

charleskorn (approved):
🚀

PR #284, opened 2025-09-05
Reuse DONT_FIND in Makefile

andyasp:
https://github.com/grafana/rollout-operator/pull/283 introduced another DONT_FIND in the Makefile, but the existing one can be reused instead.


duricanikolic (approved):
LGTM

PR #288, opened 2025-09-11
Move dashboard mixin to rollout-operator repo

tcp13equals2:
This PR brings in a Grafana dashboard which can be used for the rollout-operator. 

This dashboard can be referenced and bundled in with other dashboard mixins (such as those provided in mimir), or it can be used to create a standalone rollout-operator dashboard.

### DIFF
--- a/operations/rollout-operator-mixin/config.libsonnet
+++ b/operations/rollout-operator-mixin/config.libsonnet
@@ -18,9 +18,6 @@
     rollout_operator_instance_matcher:
       if $._config.helm == '' then $._config.rollout_operator_container_name + '.*' else '(.*%g-)?%g.*' % [$._config.helm, $._config.rollout_operator_container_name],
 
-    // The prefix including product name used when building dashboards.
-    dashboard_prefix: '%(product)s / ' % $._config.product,
-
     // Shared crosshair, the crosshair will appear on all panels but the
     graph_tooltip: 1,
 
--- a/operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
+++ b/operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
@@ -32,7 +32,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
 
   dashboard(title)::
     super.dashboard(
-      title='%(prefix)s%(title)s' % { prefix: $._config.dashboard_prefix, title: title },
+      title=$._config.rollout_operator_name,
       datasource=$._config.dashboard_datasource,
       datasource_regex=$._config.datasource_regex
     ) + {
@@ -51,18 +51,6 @@ local utils = import 'mixin-utils/utils.libsonnet';
       addClusterSelectorTemplates()::
         local d = self {
           tags: $._config.tags,
-          links: [
-            {
-              asDropdown: true,
-              icon: 'external link',
-              includeVars: true,
-              keepTime: true,
-              tags: $._config.tags,
-              targetBlank: false,
-              title: if $._config.product == '' then $._config.rollout_operator_name + ' dashboard' else $._config.product + 's dashboards',
-              type: 'dashboards',
-            },
-          ],
         };
 
         d
--- a/operations/rollout-operator-mixin-compiled/dashboards/rollout-operator.json
+++ b/operations/rollout-operator-mixin-compiled/dashboards/rollout-operator.json
@@ -14,18 +14,7 @@
       "gnetId": null,
       "graphTooltip": 1,
       "hideControls": false,
-      "links": [
-         {
-            "asDropdown": true,
-            "icon": "external link",
-            "includeVars": true,
-            "keepTime": true,
-            "tags": [ ],
-            "targetBlank": false,
-            "title": "rollout-operator dashboard",
-            "type": "dashboards"
-         }
-      ],
+      "links": [ ],
       "refresh": "5m",
       "rows": [
          {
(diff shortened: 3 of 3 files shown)


operations/rollout-operator-mixin-tools/serve/provisioning-datasources.yaml
```diff
@@ -0,0 +1,33 @@
...
+      basicAuthPassword: $DATASOURCE_PASSWORD
+    version: 1
+    editable: true
+  - name: Mimir (ingest-storage local dev env)
+    type: prometheus
+    access: proxy
+    orgId: 1
+    url: http://nginx:8080/prometheus
+    version: 1
+    editable: true
```
charleskorn: Is this needed? Seems to duplicate the datasource above.

operations/rollout-operator-mixin/config.libsonnet
```diff
@@ -0,0 +1,63 @@
...
+    rollout_operator_container_name: $._config.rollout_operator_name,
+    rollout_operator_instance_matcher:
+      if $._config.helm == '' then $._config.rollout_operator_container_name + '.*' else '(.*%g-)?%g.*' % [$._config.helm, $._config.rollout_operator_container_name],
+
+    // The prefix including product name used when building dashboards.
+    dashboard_prefix: '%(product)s / ' % $._config.product,
+
+    // If Mimir is deployed as a single binary, set to  to
+    // modify the job selectors in the dashboard queries.
+    singleBinary: false,
```
charleskorn: Is this relevant for rollout-operator?

operations/rollout-operator-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,4 @@
+{
+  grafanaDashboards+:
+    (import 'dashboards/rollout-operator.libsonnet')($._config) + { _config:: $._config },
```
charleskorn: This seems weird - why is `$._config` passed as a parameter and then merged over the top of the result? 

I would expect that what is done in `mixin.libsonnet` is sufficient and there is no need to mention `$._config` here at all.

operations/rollout-operator-mixin/config.libsonnet
tcp13equals2: That is a question I had for you ... so the variable name is a bit misleading in a non mimir context.

If singleBinary=true, then the dashboard is generated with a job variable - and the multi-select list is seeded with a list of all the rollout-operator jobs found across all the available clusters. 

If singleBinary=false, then the dashboard is generated with variables for selecting the cluster and namespace.

* does it make sense to support both? 
* should we name this variable to something else?

charleskorn: I think only supporting the cluster and namespace option is fine, we can always add the other one later if we need it.

operations/rollout-operator-mixin/mixin.libsonnet
```diff
@@ -0,0 +1,6 @@
+(import 'config.libsonnet') + {
+  _config+: {
+    product: 'Mimir',
+  },
+} +
```
charleskorn: Should this be present in this repo?

tcp13equals2: This is here to be able to generate a compiled dashboard. It serves as a working dashboard which could be used with mimir.

charleskorn: Ah, in that case, I would have a separate config file used to generate the compiled version (or perhaps we could set `dashboard_prefix` to an empty string if `product` is not set?)

tcp13equals2: I've updated to support the empty product string.

operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -0,0 +1,257 @@
...
+              asDropdown: true,
+              icon: 'external link',
+              includeVars: true,
+              keepTime: true,
+              tags: $._config.tags,
+              targetBlank: false,
+              title: if $._config.product == '' then $._config.rollout_operator_name + ' dashboard' else $._config.product + 's dashboards',
+              type: 'dashboards',
+            },
+          ],
```
charleskorn: This is not necessary here - this adds the "Mimir dashboards" link in the top-right of the dashboard, but this doesn't make sense in the context of the standalone rollout-operator dashboard.

When we incorporate this into the Mimir mixin, we should add this menu back in there.

operations/rollout-operator-mixin/config.libsonnet
```diff
@@ -0,0 +1,62 @@
...
+
+    // the name for the rollout-operator. This is also used as the container name
+    rollout_operator_name: 'rollout-operator',
+
+    rollout_operator_container_name: $._config.rollout_operator_name,
+    rollout_operator_instance_matcher:
+      if $._config.helm == '' then $._config.rollout_operator_container_name + '.*' else '(.*%g-)?%g.*' % [$._config.helm, $._config.rollout_operator_container_name],
+
+    // The prefix including product name used when building dashboards.
+    dashboard_prefix: '%(product)s / ' % $._config.product,
```
charleskorn: Should this be empty if `product` is not set? Otherwise the dashboard title renders as ` / rollout-operator`.

I think it'd be fine to drop this entirely and just make the dashboard title always "rollout-operator".

tcp13equals2: Yes, sorry. I missed that one!

I have updated per your suggestion.

PR #299, opened 2025-09-25
Refactor how dashboard is generated to avoid late binding issues

tcp13equals2:
This PR changes the way that the rollout-operator dashboard is generated.

Having the dashboard instantiated with a config means that when vendored the right most config is not honoured.

To adjust for this the [filename] label has been made static.

### DIFF
--- a/operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
+++ b/operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
@@ -27,8 +27,6 @@ builder {
 
   local sortAscending = 1,
 
-  _config:: error 'must provide _config',
-
   row(title)::
     super.row(title),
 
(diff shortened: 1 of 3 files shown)


operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -27,8 +27,6 @@ builder {
 
   local sortAscending = 1,
 
-  _config:: error 'must provide _config',
-
```
charleskorn, line 31: Why does this need to be removed? Isn't it still overridden with the value from `rollout-operator.libsonnet` line 5?

tcp13equals2: You would think so ... but with this left in it will fail. The $_config references are successful in the dashboard::

PR #300, opened 2025-09-26
Change dashboard generation

tcp13equals2:
This PR changes how the rollout-operator dashboard and utils jsonnet were setup. 

The changes made here are to better support this rollout-operator-mixin from being vendored in.

Matching PRs which vendor this in are;

* https://github.com/grafana/mimir/pull/12688
* https://github.com/grafana/deployment_tools/pull/349215


operations/rollout-operator-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -1,10 +1,10 @@
 local utils = import 'mixin-utils/utils.libsonnet';
 
-// This is a modified version of the mimir-mixin/dashboard-utils.libsonnet
+// This is a modified version of the mimir-mixin/dashboard-utils.libsonnet.
+// Note that some fields/functions have a rop_ prefix. This has been added when the same name exists in other dashboard utils with a different signature.
```
charleskorn: [nit] `rop` isn't a prefix or term we use anywhere else, so the meaning might not be immediately obvious - what if we used something like `rolloutOperator`?

charleskorn (approved):

PR #312, opened 2025-10-10
Ensure that the eviction cache keeps the pod until it's phase changes

tcp13equals2:
This PR addresses a situation where concurrent cross zone evictions can be allowed if the pod observer receives a pod update where the phase is still Running but an underlying container state is no longer good.

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/ 
`
Running: The Pod has been bound to a node, and all of the containers have been created. At least one container is still running, or is in the process of starting or restarting.
`

This change causes the pod to be evicted from the lingering eviction cache, and since the pod is no longer included in the StatefulSet tallies the zone is considered healthy.

This change ensures that the pod will remain in the eviction cache until the entire phase changes or the cache TTL is reached.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## main / unreleased
 
 * [ENHANCEMENT] Add further information to debug logs emitted when processing ZonedPodDisruptionBudget-related requests. #311 
-* [BUGFIX] Keep evicted pod in eviction cache until its phase changes from running, rather than evicting on a pod ContainerStatuses change. #312
+* [BUGFIX] Keep evicted pod in eviction cache until its phase changes from running, rather than evicting on a pod ContainerStatuses change. Increase pod eviction cache TTL. #312
 
 ## v0.30.0
 


pkg/zpdb/pod_observer.go
```diff
@@ -81,11 +79,12 @@ func (c *podObserver) invalidatePodEvictionCache(obj interface{}, action string)
 		return
 	}
 
-	// after an eviction request is allowed, the informer observes a pod update which can show it still ready/running
+	// after an eviction request is allowed, the informer observes one or more pod updates which can show it still running
 	// if another pod eviction request comes in before the first eviction takes effect this can incorrectly allow this later eviction request to proceed
-	if util.IsPodRunningAndReady(pod) {
+	// keep the cached eviction until we observe a non-running phase or the record is expired
+	if pod.Status.Phase == corev1.PodRunning {
```
charleskorn, line 85: Do we need to check both `pod.Status.Phase` and the individual container states? Or does `pod.Status.Phase` only change out of `Running` once all containers have stopped?

What does Kubernetes' PDB implementation do here?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## main / unreleased
 
 * [ENHANCEMENT] Add further information to debug logs emitted when processing ZonedPodDisruptionBudget-related requests. #311 
+* [BUGFIX] Keep evicted pod in eviction cache until it's phase changes from running. #312
```
charleskorn:
```suggestion
* [BUGFIX] Keep evicted pod in eviction cache until its phase changes from running, rather than only waiting for each container in the pod to stop running. #312
```

pkg/zpdb/pod_observer.go
tcp13equals2: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

The Pod will be reporting Running until all the containers have stopped.

tcp13equals2: The existing `IsReadyAndRunning()` checks that both the Phase == PodRunning and all the containers are running. Which is good for testing positive health. 

But using this for the eviction cache means we can remove the pod from the cache whilst it's phase is still PodRunning

CHANGELOG.md
tcp13equals2: How about;

[BUGFIX] Keep evicted pod in eviction cache until its phase changes from running, rather than evicting on a pod ContainerStatuses change. #312

charleskorn: Sounds good to me

PR #324, opened 2025-10-22
Update rollout-operator controller to use zpdb before delete

tcp13equals2:
This PR seeks to address a race condition which can occur when rolling pod updates are running and a voluntary pod eviction is received.

During the main `reconcileStatefulSetsGroup()` reconcile loop, there is opportunity for a voluntary eviction to have occurred, but the reconcile loop has not detected this and a pod can be deleted (for updating) which would breach the ZPDB.

This is more likely to occur when running the zone aware PodDisruptionBudget in a partition awareness mode. In the traditional max unavailable = 1 PDB, a voluntary eviction is denied if there is any disruption in any zone. Whilst the rolling update controller is performing updates the likelihood of a a voluntary eviction being allowed is small.

However, when in partition awareness mode, the voluntary eviction will be allowed if both pods in the partition are ready. ie a reconcile over `ingester-zone-a` pods is occurring - with each pod being deleted, a voluntary eviction comes in for `ingester-zone-b-50`. This eviction will be allowed if the reconcile loop has not yet reached `ingester-zone-a-50`. The update loop will not be aware of this eviction and issue a delete on `ingester-zone-a-50`.

The fix for this is to call the zpdb eviction controller prior to each delete and confirm that the zpdb will not be breached by this delete. 

Note - this will result in additional statefulset and pod checks before each rolling pod update (delete).

### DIFF
--- a/pkg/zpdb/eviction_controller.go
+++ b/pkg/zpdb/eviction_controller.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/grafana/rollout-operator/pkg/controller"
 	"github.com/grafana/rollout-operator/pkg/util"
 )
 
@@ -28,8 +27,6 @@ const (
 	logDenyMesg  = "pod eviction denied"
 	logAllowMesg = "pod eviction allowed"
 )
-
-var _ controller.ZPDBEvictionController = (*EvictionController)(nil)
 
 type EvictionController struct {
 	// a lock used to control finding a specific named lock


pkg/controller/controller.go
```diff
@@ -575,9 +579,32 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
 				continue
 			}
 
-			level.Info(c.logger).Log("msg", fmt.Sprintf("terminating pod %s", pod.Name))
-			if err := c.kubeClient.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
-				return false, errors.Wrapf(err, "failed to delete pod %s", pod.Name)
+			if c.zpdbController != nil && c.zpdbController.Running() {
```
charleskorn: When would `zpdbController` ever be `nil`, or not be running when we get here?

And if it's not running, should we stop? Otherwise we might stop a pod that would otherwise not be allowed.

pkg/zpdb/eviction_controller.go
```diff
@@ -93,6 +106,51 @@ func (c *EvictionController) findLock(name string) *sync.Mutex {
 	return c.locks[name]
 }
 
+// AllowPodEvictionRequest allows for a programmatic pod eviction request. It returns an error if the pod eviction is denied.
+// Note that if this func returns without an error, the pod will be marked as pending eviction within the zpdb eviction cache.
+// Note also that this func does not actually evict or delete the pod from kubernetes.
+func (c *EvictionController) AllowPodEvictionRequest(ctx context.Context, namespace string, podname string, source string) error {
```
charleskorn: [nit] Maybe a name like `MarkPodAsDeleted` would be clearer here?

pkg/controller/controller.go
```diff
@@ -575,9 +579,32 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
...
+				err := c.zpdbController.AllowPodEvictionRequest(ctx, pod.Namespace, pod.Name, "rollout-controller")
+				if err != nil {
+					// Skip this pod. The reconcile loop regularly runs and this pod will have an opportunity to be re-tried.
+					// Rather than returning false here and abandoning the rest of the updates until the next reconcile,
+					// we allow this update loop to continue. For configurations which have a partition awareness ZPDB it is valid
+					// to have multiple disruptions as long as there is at least one health pod per partition.
+					level.Debug(c.logger).Log("msg", fmt.Sprintf("zpdb denied pod deletion. pod=%s, reason=%s", pod.Name, err.Error()))
+					continue
+				}
+				level.Info(c.logger).Log("msg", fmt.Sprintf("terminating pod %s. zpdb allowed", pod.Name))
```
charleskorn:
```suggestion
				level.Info(c.logger).Log("msg", "terminating pod (does not violate any relevant ZPDBs)", "pod", pod.Name)
```

pkg/controller/controller.go
tcp13equals2: Currently, the zpdb eviction controller only runs if the rollout-operator is running with webhooks enabled.

We could run the eviction controller all the time and always vector these controller deletes via it.

Do you think this is a better way to go?

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -134,19 +133,16 @@
...
           policyRule.withApiGroups($.replica_template.spec.group) +
           policyRule.withResources(['%s/scale' % $.replica_template.spec.names.plural, '%s/status' % $.replica_template.spec.names.plural]) +
           policyRule.withVerbs(['get', 'patch']),
         ] else []
-      ) + (
-        if enableWebhooks then [
-          policyRule.withApiGroups($.zpdb_template.spec.group) +
-          policyRule.withResources([$.zpdb_template.spec.names.plural]) +
-          policyRule.withVerbs(['get', 'list', 'watch']),
-        ] else []
```
charleskorn: In the interests of avoiding churn in the role definition (even though what you've done is unquestionably clearer), should we leave the ZPDB-related permissions at the end here?

pkg/controller/controller.go
```diff
@@ -575,7 +579,24 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
...
+			// For each request, a full set of tests is performed to confirm the state of all pods in the ZPDB scope.
+			// This ensures that if any voluntary evictions have been allowed since this reconcile loop commenced that
+			// the latest information is used to determine the zone/partition disruption level.
+			// If this request returns without error, the pod will be placed into the ZPDB pod eviction cache and will
+			// be considered as not ready until it either restarts or is expired from the cache.
+			err := c.zpdbController.MarkPodAsDeleted(ctx, pod.Namespace, pod.Name, "rollout-controller")
+			if err != nil {
+				// Skip this pod. The reconcile loop regularly runs and this pod will have an opportunity to be re-tried.
+				// Rather than returning false here and abandoning the rest of the updates until the next reconcile,
+				// we allow this update loop to continue. For configurations which have a partition awareness ZPDB it is valid
```
charleskorn:
```suggestion
				// we allow this update loop to continue. For configurations which have a partition aware ZPDB it is valid
```

pkg/controller/controller.go
```diff
@@ -575,7 +579,24 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
...
+				// to have multiple disruptions as long as there is at least one health pod per partition.
```
charleskorn:
```suggestion
				// to have multiple disruptions as long as there is at least one healthy pod per partition.
```

pkg/controller/controller.go
```diff
@@ -575,7 +579,24 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
...
+				level.Debug(c.logger).Log("msg", fmt.Sprintf("zpdb denied pod deletion. pod=%s, reason=%s", pod.Name, err.Error()))
```
charleskorn: There's no need to format strings like this for log messages, we emit structured key-value pairs:

```suggestion
				level.Debug(c.logger).Log("msg", "zpdb denied pod deletion", "pod", pod.Name, "reason", err)
```

pkg/controller/controller_test.go
charleskorn, line 1: Have you considered mocking the ZPDB `EvictionController` for these tests? It might help make them a little simpler and make it easier to test different scenarios.

charleskorn, line 1: It'd be good to see a test here for the scenario that prompted this PR.

pkg/zpdb/eviction_controller.go
```diff
@@ -58,17 +65,32 @@ func NewEvictionController(kubeClient kubernetes.Interface, dynamicClient dynami
...
 	}
 	if err := c.podObserver.start(); err != nil {
 		return errors.Wrap(err, "failed to start zpdb pod observer")
 	}
+	c.running = true
 	return nil
 }
 
+// Running returns true if this controller has been successfully started and not yet stopped
+func (c *EvictionController) Running() bool {
```
charleskorn: Is this still needed?

pkg/zpdb/eviction_controller.go
```diff
@@ -35,6 +41,7 @@ type EvictionController struct {
 
 	cfgObserver *configObserver
 	podObserver *podObserver
+	running     bool
```
charleskorn: Is this still needed?

pkg/zpdb/eviction_controller.go
```diff
@@ -58,17 +65,32 @@ func NewEvictionController(kubeClient kubernetes.Interface, dynamicClient dynami
 	}
 }
 
+// AddOrUpdateConfig attempts to parse and apply the given unstructured object as a zpdb Config
+// An error is returned if the given object can not be parsed into a valid config.
+// The bool response indicates if the config was accepted into the configCache, with the int64 reflecting the generation of the config in the configCache.
+// It will return false if this given object is an older generation of what is already in the configCache.
+func (c *EvictionController) AddOrUpdateConfig(pdbRawConfig *unstructured.Unstructured) (bool, int64, error) {
```
charleskorn: If the `EvictionController` is mocked out for the rollout controller tests, then we shouldn't need to add this method here.

pkg/zpdb/eviction_controller.go
```diff
@@ -93,6 +115,51 @@ func (c *EvictionController) findLock(name string) *sync.Mutex {
 	return c.locks[name]
 }
 
+// MarkPodAsDeleted allows for a programmatic pod eviction request. It returns an error if the pod eviction is denied.
+// Note that if this func returns without an error, the pod will be marked as pending eviction within the zpdb eviction cache.
+// Note also that this func does not actually evict or delete the pod from kubernetes.
+func (c *EvictionController) MarkPodAsDeleted(ctx context.Context, namespace string, podname string, source string) error {
```
charleskorn: [nit]

```suggestion
func (c *EvictionController) MarkPodAsDeleted(ctx context.Context, namespace string, podName string, source string) error {
```

pkg/zpdb/eviction_controller.go
```diff
@@ -93,6 +115,51 @@ func (c *EvictionController) findLock(name string) *sync.Mutex {
...
+	request := v1.AdmissionReview{
+		Request: &v1.AdmissionRequest{
+			// not a real uid. The eviction_controller only uses this for logging purposes
+			UID: types.UID(fmt.Sprintf("request-pod-eviction-%s", podname)),
```
charleskorn: Might be good to make it super clear that this is not a real eviction request, for example:
```suggestion
			UID: types.UID(fmt.Sprintf("internal-request-pod-eviction-%s-for-rolling-update", podname)),
```

pkg/zpdb/eviction_controller.go
```diff
@@ -23,6 +28,12 @@ const (
 	logAllowMesg = "pod eviction allowed"
 )
 
+type IEvictionController interface {
+	MarkPodAsDeleted(ctx context.Context, namespace string, podName string, source string) error
+}
+
+var _ IEvictionController = (*EvictionController)(nil)
```
charleskorn: The convention for interfaces like this that are used as a way to break a dependency on a concrete type is to define them in the consuming package, unless they're widely used.

So I'd move this interface to `pkg/controller` and call it something like `ZPDBEvictionController`.

pkg/zpdb/eviction_controller_test.go
```diff
@@ -120,6 +135,17 @@ func (c *testContext) assertDenyResponse(t *testing.T, reason string, statusCode
...
+	response := c.controller.MarkPodAsDeleted(c.ctx, testNamespace, pod, "eviction-controller-test")
+	require.ErrorContains(t, response, reason)
+	c.stop()
+}
+
+func (c *testContext) assertAllowResponseViaMarkPodAsDeleted_noStop(t *testing.T, pod string) {
+	response := c.controller.MarkPodAsDeleted(t.Context(), testNamespace, pod, "eviction-controller-test")
+	require.NoError(t, response)
+}
+
```
charleskorn, line 142: [nit] It's a bit unexpected to me that a method called `assertDenyResponseViaMarkPodAsDeleted` also stops the controller - I would make this explicit in the tests themselves and then the `assertAllow...` method doesn't need the `noStop` suffix.

pkg/controller/controller_test.go
```diff
@@ -1254,3 +1291,18 @@ func (f *fakeHttpClient) requests() []string {
 
 	return f.recordedRequests
 }
+
+var _ zpdb.IEvictionController = (*mockEvictionController)(nil)
```
charleskorn: This kind of assertion isn't needed - passing `mockEvictionController` to `NewRolloutController` will fail with a clear compilation error if `mockEvictionController` doesn't satisfy the interface.

pkg/zpdb/eviction_controller.go
```diff
@@ -23,6 +29,8 @@ const (
 	logAllowMesg = "pod eviction allowed"
 )
 
+var _ controller.ZPDBEvictionController = (*EvictionController)(nil)
```
charleskorn: I wouldn't bother with this assertion here - if we break the expectation that `EvictionController` satisfies the `controller.ZPDBEvictionController` interface, then rollout-operator as a whole will still fail to compile, and removing this assertion also removes a dependency between the two packages.

charleskorn (approved):

PR #334, opened 2025-11-14
Adding metrics to zpdb

tcp13equals2:
This PR adds some metrics to report on internal ZPDB state.

This includes;
* rollout_operator_zpdb_configurations_observed_total
* rollout_operator_zpdb_eviction_requests_total
* rollout_operator_zpdb_inflight_eviction_requests_total
* kube_validating_webhook_failure_policy

rollout_operator_zpdb_configurations_observed_total - with the use of labels we can identify successfully applied configurations as well as invalid or ignored configurations

rollout_operator_zpdb_eviction_requests_total - with the use of labels we can track all accept, deny and error results in eviction requests

rollout_operator_zpdb_inflight_eviction_requests_total - tracks the number of in-flight (queued) requests

kube_validating_webhook_failure_policy - tracks the FailurePolicy setting on each webhook.

A subsequent PR will add dashboard panels to leverage these metrics.

Example metrics output;

```
kube_validating_webhook_failure_policy{policy="Fail",type="MutatingWebhook",webhook="prepare-downscale-rollout-operator-development.grafana.com"} 1
kube_validating_webhook_failure_policy{policy="Fail",type="ValidatingWebhook",webhook="no-downscale-rollout-operator-development.grafana.com"} 1
kube_validating_webhook_failure_policy{policy="Fail",type="ValidatingWebhook",webhook="pod-eviction-rollout-operator-development.grafana.com"} 1
kube_validating_webhook_failure_policy{policy="Ignore",type="ValidatingWebhook",webhook="zpdb-validation-rollout-operator-development.grafana.com"} 1

```

### DIFF
--- a/pkg/webhooks/webhook_collector_test.go
+++ b/pkg/webhooks/webhook_collector_test.go
@@ -77,12 +77,6 @@ func TestWebhookCollector_CollectsWebhookFailurePolicies(t *testing.T) {
 	)
 	require.NoError(t, err)
 
-	// Wait until the collector has observed and cached all three inner webhooks
-	expectedSeries := 6
-	require.Eventually(t, func() bool {
-		return testutil.CollectAndCount(collector, "kube_validating_webhook_failure_policy") == expectedSeries
-	}, 5*time.Second, 10*time.Millisecond, "expected %d samples from collector", expectedSeries)
-
 	// Verify metric samples and labels
 	expected := `
 	# HELP kube_validating_webhook_failure_policy FailurePolicy setting of Kubernetes ValidatingWebhooks and MutatingWebhooks
@@ -94,8 +88,10 @@ func TestWebhookCollector_CollectsWebhookFailurePolicies(t *testing.T) {
 	kube_validating_webhook_failure_policy{policy="Fail",type="MutatingWebhook",webhook="mutating-webhook-with-fail"} 1
 	kube_validating_webhook_failure_policy{policy="Ignore",type="MutatingWebhook",webhook="mutating-webhook-with-fail"} 0
 	`
-	err = testutil.CollectAndCompare(collector, bytes.NewBufferString(expected), "kube_validating_webhook_failure_policy")
-	require.NoError(t, err)
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		err = testutil.CollectAndCompare(collector, bytes.NewBufferString(expected), "kube_validating_webhook_failure_policy")
+		require.NoError(t, err)
+	}, 5*time.Second, 10*time.Millisecond)
 
 	// Update the policy to Ignore and await the collector to report this in the metrics
 	mutatingConfiguration.Webhooks[0].FailurePolicy = &ignore


cmd/rollout-operator/main.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/grafana/rollout-operator/pkg/controller"
 	"github.com/grafana/rollout-operator/pkg/instrumentation"
 	"github.com/grafana/rollout-operator/pkg/tlscert"
-	"github.com/grafana/rollout-operator/pkg/zpdb"
+	zpdb "github.com/grafana/rollout-operator/pkg/zpdb"
```
charleskorn: Why is this change needed?

pkg/zpdb/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+	EvictionRequests       *prometheus.CounterVec
+	InFlightRequests       *prometheus.GaugeVec
+}
+
+func NewMetrics(reg prometheus.Registerer) *Metrics {
+	return &Metrics{
+		ConfigurationsObserved: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "rollout_operator_zpdb_configurations_observed_total",
+			Help: "Number of zpdb configurations observed by the configuration controller.",
+		}, []string{"result", "name"}),
```
charleskorn: Including the ZPDB name as a label is going to increase the cardinality significantly - is there a reason behind including it?

pkg/zpdb/config_observer.go
```diff
@@ -84,16 +86,20 @@ func (c *configObserver) addOrUpdate(obj *unstructured.Unstructured) {
...
 	} else {
 		level.Info(c.logger).Log("msg", "zpdb configuration update ignored", "name", obj.GetName(), "generation", generation, "ignored-generation", obj.GetGeneration())
+		c.metrics.ConfigurationsObserved.WithLabelValues("ignored", obj.GetName()).Inc()
 	}
 }
 
 func (c *configObserver) onPdbAdded(obj interface{}) {
 	unstructuredObj, isUnstructured := obj.(*unstructured.Unstructured)
 	if !isUnstructured {
+		c.metrics.ConfigurationsObserved.WithLabelValues("not-unstructured", "unknown").Inc()
```
charleskorn: Given we don't expect to receive values that aren't `Unstructured` instances, what if we group these together with `invalid` rather than introducing a special label value for it?

(same applies below)

pkg/zpdb/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+		EvictionRequests: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "rollout_operator_zpdb_eviction_requests_total",
+			Help: "Number of zpdb eviction requests.",
+		}, []string{"reason", "pod", "status"}),
+		InFlightRequests: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "rollout_operator_zpdb_inflight_eviction_requests_total",
+			Help: "Number of zpdb eviction requests which are currently in-flight.",
+		}, []string{"pod"}),
```
charleskorn: Including the pod name as a label for these metrics is going to increase their cardinality significantly - is there a reason behind including it?

charleskorn: It looks like we already have equivalent metrics in place (`rollout_operator_request_duration_seconds_count` and `rollout_operator_inflight_requests`, both of which can be filtered by route to ZPDB-related routes) - am I missing something? Is there something different about these metrics?

pkg/zpdb/metrics.go
tcp13equals2: You don't think having this name is worth the extra cardinality? 

I was thinking having this as a label would allow for better dashboard panels to be constructed - as a specific zpdb could be referenced or grouped by. Similarly for alerting.

pkg/zpdb/metrics.go
tcp13equals2: I included pod name as I thought it would be an interesting one to be able to configure dashboards and alerts on.

It would be interesting to see if there are specific pods being evicted more then others. The pod name would also give context to a high inflight request count. 

With the pod name for ingesters/storegateway we could also be able to aggregate on the container or container + zone - which again would be interesting to see which are being evicted compared to the others.

Monitoring for these voluntary evictions might give insight into kubernetes relieving node pressure, drains, etc

tcp13equals2: The existing metrics are wired as middleware on the http handler. These will not include the zpdb invocations which have come directly from the rollout controller.  The existing metrics would also not have the pod name as a label. Per comment above.

pkg/zpdb/metrics.go
charleskorn: My 2c is no, but I don't hold this opinion strongly. Dashboards would give us a high-level overview, and logs can give us the low-level detail of which ZPDB is failing and why.

pkg/zpdb/metrics.go
charleskorn: Right, makes sense, thanks for the explanation.

charleskorn: Do you have an example of the kind of alert you're imagining would need the `pod` label?

tcp13equals2: The pod label lets us work backwards to the container and container + zone. If we every want an alert on number of evictions, this gives us something to group by. ie alert based off the number of ingester evictions vs ingester-zone-a evictions.

I think that with multi-az deployments, the zone becomes more interesting.

Would you be more comfortable with using labels such as container and zone rather then the full pod name? Or just the STS name would also allow the container + zone to be identified. 

I agree that we probably don't need the full pod name for the scenarios I have commented here, but the STS name could be useful.

charleskorn: StatefulSet name sounds like a good compromise to me.

pkg/webhooks/webhook_collector.go
charleskorn, line 1: Could you please add some tests for this?

tcp13equals2: I've added some tests for this collector.

pkg/webhooks/webhook_collector_test.go
charleskorn, line 1: The corresponding non-test file is called `webhook_collector.go` - could you please rename this file to match?

pkg/webhooks/webhooks_collector_test.go
```diff
@@ -0,0 +1,136 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+)
+
+func TestStartStop(t *testing.T) {
+	client := k8sfake.NewClientset()
+	collector := NewWebhookCollector(client, "test", log.NewNopLogger())
+	require.NoError(t, collector.Start())
+	collector.Stop()
+}
```
charleskorn: This test seems unnecessary given the test below has exactly the same assertions.

pkg/webhooks/webhook_collector.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (c *WebhookCollector) Collect(ch chan<- prometheus.Metric) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	for name, failureMode := range c.validatingWebhooksSettings {
+		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, 1, "ValidatingWebhook", name, string(failureMode))
+	}
+	for name, failureMode := range c.mutatingWebhooksSettings {
+		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, 1, "MutatingWebhook", name, string(failureMode))
+	}
```
charleskorn, line 101: It might be a good idea to emit 0 values for the other failure mode value - otherwise queries within the lookback period of a change will show both failure modes set.

pkg/webhooks/webhook_collector_test.go
```diff
@@ -0,0 +1,138 @@
...
+	kube_validating_webhook_failure_policy{policy="Fail",type="ValidatingWebhook",webhook="validating-webhook-with-ignore"} 0
+	kube_validating_webhook_failure_policy{policy="Fail",type="ValidatingWebhook",webhook="no-failure-policy-set"} 1
+	kube_validating_webhook_failure_policy{policy="Ignore",type="ValidatingWebhook",webhook="no-failure-policy-set"} 0
+	kube_validating_webhook_failure_policy{policy="Ignore",type="MutatingWebhook",webhook="mutating-webhook-with-fail"} 1
+	kube_validating_webhook_failure_policy{policy="Fail",type="MutatingWebhook",webhook="mutating-webhook-with-fail"} 0
+	`
+	require.Eventually(t, func() bool {
+		err := testutil.CollectAndCompare(collector, bytes.NewBufferString(expected), "kube_validating_webhook_failure_policy")
+		return err == nil
+	}, 5*time.Second, 10*time.Millisecond, "awaiting MutatingWebhook to have policy=Ignore")
```
charleskorn: Have you considered using `EventuallyWithT` here? It might help produce a clearer error message when this fails.

pkg/webhooks/webhook_collector_test.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+		metav1.CreateOptions{},
+	)
+	require.NoError(t, err)
+
+	// Wait until the collector has observed and cached all three inner webhooks
+	expectedSeries := 6
+	require.Eventually(t, func() bool {
+		return testutil.CollectAndCount(collector, "kube_validating_webhook_failure_policy") == expectedSeries
+	}, 5*time.Second, 10*time.Millisecond, "expected %d samples from collector", expectedSeries)
```
charleskorn: What if we wrapped the `testutil.CollectAndCompare` immediately below this in a `require.EventuallyWithT` call? Then this would not be necessary, and we'd get a clearer assertion error message if it fails.

pkg/webhooks/webhook_collector.go
```diff
@@ -0,0 +1,106 @@
...
+		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, 0, "ValidatingWebhook", name, string(c.opposite(failureMode)))
+	}
+	for name, failureMode := range c.mutatingWebhooksSettings {
+		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, 1, "MutatingWebhook", name, string(failureMode))
+		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, 0, "MutatingWebhook", name, string(c.opposite(failureMode)))
+	}
+}
+
+// opposite returns the opposite FailurePolicyType for the given policy type
+func (c *WebhookCollector) opposite(mode admissionregistrationv1.FailurePolicyType) admissionregistrationv1.FailurePolicyType {
```
charleskorn: This might behave unexpectedly if there's ever another value for `FailurePolicyType` introduced. What if we change `Collect` above to something along these lines?

```go
	for name, failureMode := range c.validatingWebhooksSettings {
		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, boolToFloat(failureMode == admissionregistrationv1.Ignore), "ValidatingWebhook", name, "Ignore")
		ch <- prometheus.MustNewConstMetric(c.metricDescription, prometheus.GaugeValue, boolToFloat(failureMode == admissionregistrationv1.Fail), "ValidatingWebhook", name, "Fail")
	}
```

Then at least the series emitted will be correct, even if they're all 0s.

pkg/webhooks/webhook_collector_test.go
charleskorn: Looks like this was missed @tcp13equals2

tcp13equals2: @charleskorn have a look now ...

charleskorn (approved):

PR #343, opened 2025-12-03
feat: Handle 409 status code returned by `DELETE prepared_scale_down`

Segflow:
The 409 Conflict status code from `DELETE /{prepared_scale_down}` should trigger failure on the entire error group - rollout operator should suppress this error and continue with the other endpoints.

### DIFF
--- a/pkg/controller/delay.go
+++ b/pkg/controller/delay.go
@@ -203,7 +203,7 @@ func callPrepareDownscaleAndReturnElapsedDurationsSinceInitiatedDownscale(ctx co
 				// Unlike `callCancelDelayedDownscale`, here we consider all non 2xx status code as error and should block the downscale.
 				// this includes the case where we fail because a partition state change is locked.
 				level.Error(epLogger).Log("msg", "unexpected status code returned when calling POST on endpoint", "status", resp.StatusCode, "response_body", string(body))
-				return fmt.Errorf("HTTP DELETE request returned non-2xx status code: %v", resp.StatusCode)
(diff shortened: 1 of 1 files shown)


pkg/controller/delay.go
```diff
@@ -260,8 +260,15 @@ func callCancelDelayedDownscale(ctx context.Context, logger log.Logger, client h
 			defer resp.Body.Close()
 
 			if resp.StatusCode/100 != 2 {
-				err := errors.New("HTTP DELETE request returned non-2xx status code")
 				body, readError := io.ReadAll(resp.Body)
+
+				// Handle 409 Conflict separately - this typically means partition state is locked by an operator
```
salvacorts: Can you share an example when a 409 would be returned?

salvacorts: I see it here https://github.com/grafana/loki/pull/20141

pkg/controller/delay.go
```diff
@@ -260,8 +260,15 @@ func callCancelDelayedDownscale(ctx context.Context, logger log.Logger, client h
...
+				if resp.StatusCode == http.StatusConflict {
```
pracucci, line 268: It's correct not apply this logic to `callPrepareDownscaleAndReturnElapsedDurationsSinceInitiatedDownscale()` because we want to block the download in that case, but we need to make sure we'll not introduce a regression in the future. I suggest to add a comment to `callPrepareDownscaleAndReturnElapsedDurationsSinceInitiatedDownscale()` to explain why we consider a 409 a blocking error there.

Segflow: commend added

PR #347, opened 2025-12-09
Webhook metrics collector apply namespace label filter to webhooks considered

tcp13equals2:


pkg/webhooks/webhook_collector_test.go
```diff
@@ -137,4 +146,29 @@ func TestWebhookCollector_CollectsWebhookFailurePolicies(t *testing.T) {
...
+					FailurePolicy: &fail,
+				},
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	require.NoError(t, err)
+
+	// The above webhook creation has been ignored - our metrics remain the same
+	time.Sleep(5 * time.Second)
```
charleskorn: Is there a more robust way to do this that doesn't involve sleeping for 5s?

tcp13equals2: Changed to a require.Never() ... is this better?

charleskorn: Is there some way for the test to detect if the new webhook has been observed? That feels more robust to me.

Maybe we could have a counter of observed webhook changes that is incremented in `OnValidatingWebhookConfiguration` and `OnMutatingWebhookConfiguration`, and then the test can wait for that to be incremented?

tcp13equals2: I have added a counter. Note that I did not expose this as a prometheus metric as it is probably not that useful externally.

charleskorn: Makes sense, I was only expecting this to be used in the tests.

PR #349, opened 2025-12-12
Additional unit tests

tcp13equals2:
This PR does not change any functionality. It only adds additional unit tests;

* tests around using additional groupings in regex (is not currently needed, but a useful test for future reference)
* test expected handling of pod updating in the rollout controller where the zpdb denies deletes
* test downscale webhook deny when another statefulset has non updated pods

No changelog has been recorded as there is no functional change.


pkg/zpdb/config_test.go
```diff
@@ -72,6 +72,28 @@ func TestPodPartitionZoneMatch(t *testing.T) {
 	}
 }
 
+func TestOptionalRegexGrouping(t *testing.T) {
```
tcp13equals2, line 75: Note to reviewers - even though we do not plan to use this regex I have left this test as it illustrates a secondary grouping.

PR #351, opened 2025-12-15
Adding a maxUnavailable override

tcp13equals2:
This PR allows the rollout controller to request a pod to be deleted even if the zpdb has a maxUnavailable=0 set.

It is plausible scenario that the operator has disabled all voluntary evictions - setting maxUnavailable=0, yet still requires the rollout controller to be scaling stateful sets up/down and updating pods.

Before the zpdb functionality was integrated into the rollout-operator, the ability for the rollout controller to delete pods was independent of the PDB. So even with maxUnavailable=0, the rollout controller could still request pod deletions.

With the integration of the zpdb into the rollout controller, this has resulted in a situation where the rollout controller can not delete (update) pods if the zpdb has a maxUnavailable=0. 

This PR introduces the ability for the rollout controller to still request for pods to be deleted even if maxUnavailable=0 is set. The zpdb eviction code will still enforce a PDB, ensuring that the pod deletion will not violate a PDB. 

To do this, the rollout controller will pass in a maxUnavailable override to the zpdb - which is only used if maxUnavailable=0.

The override which is passed in depends on the zpdb for that group of pods. If the zpdb is a partition based PDB (has a regex set to extract a partition number), then the override passed in is 1. This enforces that the partition only allows for 1 disruption. 

For classic zone zpdbs, the override passed in is the StatefulSet's maxUnavailable value. This is the same value that the rollout controller is using to set a maximum number of pod interruptions.

It is noted that at least for Mimir, in a classic zone configuration the zpdb maxUnavailable is set to the same value as the StatefulSet maxUnavailable - so this is a logical choice for the override value.

### DIFF
--- a/pkg/zpdb/eviction_controller_test.go
+++ b/pkg/zpdb/eviction_controller_test.go
@@ -418,7 +418,6 @@ func TestPodEviction_MultiZoneClassic(t *testing.T) {
 // TestPodEviction_MultiZoneClassicOverrideHasNoEffectWhenNotZero validates that the MaxUnavailableZeroOverride
 // is only used when maxUnavailable=0. Setting the override when this is not true has no effect.
 func TestPodEviction_MultiZoneClassicOverrideHasNoEffectWhenNotZero(t *testing.T) {
-
 	objs := make([]runtime.Object, 0, 12)
 	objs = append(objs, newEvictionControllerSts(statefulSetZoneA))
 	objs = append(objs, newEvictionControllerSts(statefulSetZoneB))


pkg/controller/controller.go
```diff
@@ -573,6 +575,24 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
...
 		}
 
+		// Initialize a maxUnavailable override to the ZPDB.
+		// This will only be used if the ZPDB has disabled voluntary evictions (maxUnavailable=0).
+		// Note that this will have also have no effect if the pods are not within a ZPDB scope.
+		zpdbMaxUnavailableOverrideIfZero := zpdb.NewMaxUnavailableZeroOverride(maxUnavailable)
+
+		hasPartitionAwarePdb, err := c.zpdbController.HasPartitionAwarePdb(podsToUpdate[0])
+		if err != nil {
+			// Note if ignore this error and continued processing the same error would be raised from the MarkPodAsDeleted() below.
```
aknuds1: [Nit]
```suggestion
			// Note if we ignored this error and continued processing, the same error would be raised from the MarkPodAsDeleted() below.
```

pkg/controller/controller_test.go
```diff
@@ -717,6 +737,25 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+				require.True(t, ok)
+				var stsMaxUnavailableInt int
+				if stsMaxUnavailableString == "xxx" || testData.zpdbPartitionMode {
+					stsMaxUnavailableInt = 1
+				} else {
+					stsMaxUnavailableInt, err = strconv.Atoi(stsMaxUnavailableString)
+					require.NoError(t, err)
+				}
+
+				assert.Equal(t, evictionController.lastOverride, zpdb.NewMaxUnavailableZeroOverride(stsMaxUnavailableInt))
```
aknuds1: The reference should be the second argument to `assert.Equal`:
```suggestion
				assert.Equal(t, zpdb.NewMaxUnavailableZeroOverride(expectedOverride), evictionController.lastOverride)
```

pkg/controller/controller_test.go
```diff
@@ -230,6 +232,24 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+				mockStatefulSetPod("ingester-zone-a-0", testPrevRevisionHash),
+				mockStatefulSetPod("ingester-zone-a-1", testPrevRevisionHash),
+				mockStatefulSetPod("ingester-zone-a-2", testPrevRevisionHash),
+				mockStatefulSetPod("ingester-zone-b-0", testPrevRevisionHash),
+				mockStatefulSetPod("ingester-zone-b-1", testPrevRevisionHash),
+				mockStatefulSetPod("ingester-zone-b-2", testPrevRevisionHash),
+			},
+			// the first batch of zone-a pods is deleted
+			expectedDeletedPods: []string{"ingester-zone-a-0", "ingester-zone-a-1"},
+			// setting this will trigger the maxUnavailability override if 0 to be 1
```
aknuds1: Doesn't ZPDB partition mode cause the maxUnavailability override to be _unconditionally_ 1?
```suggestion
			// setting this will trigger the maxUnavailability override to be 1
```

pkg/zpdb/eviction_controller.go
```diff
@@ -28,6 +28,20 @@ const (
 	logAllowMesg = "pod eviction allowed"
 )
 
+// MaxUnavailableZeroOverride provides an alternate maxUnavailable value to be used when the current maxUnavailable configuration is set to zero
```
aknuds1: [Nit] Idiomatic ending period:
```suggestion
// MaxUnavailableZeroOverride provides an alternate maxUnavailable value to be used when the current maxUnavailable configuration is set to zero.
```

pkg/controller/controller.go
```diff
@@ -573,6 +575,24 @@ func (c *RolloutController) updateStatefulSetPods(ctx context.Context, sts *v1.S
 			return true, nil
 		}
 
+		// Initialize a maxUnavailable override to the ZPDB.
+		// This will only be used if the ZPDB has disabled voluntary evictions (maxUnavailable=0).
+		// Note that this will have also have no effect if the pods are not within a ZPDB scope.
```
aknuds1:
```suggestion
		// Note that this will also have no effect if the pods are not within a ZPDB scope.
```

pkg/controller/controller_test.go
```diff
@@ -717,6 +737,25 @@ func TestRolloutController_Reconcile(t *testing.T) {
...
+
+				stsMaxUnavailableString, ok := sts.Annotations[config.RolloutMaxUnavailableAnnotationKey]
+				require.True(t, ok)
+				var stsMaxUnavailableInt int
+				if stsMaxUnavailableString == "xxx" || testData.zpdbPartitionMode {
+					stsMaxUnavailableInt = 1
+				} else {
+					stsMaxUnavailableInt, err = strconv.Atoi(stsMaxUnavailableString)
+					require.NoError(t, err)
+				}
```
aknuds1: We should make sure to use the StatefulSet owning the deleted pod:
```suggestion
				// Find the StatefulSet that owns the deleted pods by matching the pod name prefix.
				deletedPodName := testData.expectedDeletedPods[0]
				var sts *v1.StatefulSet
				for _, obj := range testData.statefulSets {
					s := obj.(*v1.StatefulSet)
					if strings.HasPrefix(deletedPodName, s.Name+"-") {
						sts = s
						break
					}
				}
				require.NotNil(t, sts, "could not find StatefulSet for deleted pod %s", deletedPodName)

				var expectedOverride int
				if testData.zpdbPartitionMode {
					expectedOverride = 1
				} else {
					stsMaxUnavailableString, ok := sts.Annotations[config.RolloutMaxUnavailableAnnotationKey]
					require.True(t, ok)
					if stsMaxUnavailableString == "xxx" {
						expectedOverride = 1
					} else {
						expectedOverride, err = strconv.Atoi(stsMaxUnavailableString)
						require.NoError(t, err)
					}
				}
```

pkg/zpdb/eviction_controller.go
```diff
@@ -221,6 +244,14 @@ func (c *EvictionController) HandlePodEvictionRequest(ctx context.Context, ar v1
 
 	// the number of allowed unavailable pods in other zones.
 	maxUnavailable := pdbConfig.maxUnavailablePods(sts)
+
+	// if the maxUnavailable is 0 (no voluntary evictions), replace this value with an override if it has been set.
```
aknuds1: [Nit]
```suggestion
	// if maxUnavailable is 0 (no voluntary evictions), replace this value with an override if it has been set.
```

pkg/zpdb/eviction_controller_test.go
```diff
@@ -253,7 +263,34 @@ func TestPodEviction_MaxUnavailableEq0_ViaAllowPodEviction(t *testing.T) {
...
+		objs = append(objs, newPod(p, objs[2].(*appsv1.StatefulSet)))
+	}
+
+	zoneAPod0 := objs[3].(*corev1.Pod)
+
+	testCtx := newTestContextWithoutAdmissionReview(t, newPDBMaxUnavailable(0, rolloutGroupValue), objs...)
+	defer testCtx.controller.Stop()
+	// even though the config has a maxUnavailable=0, an override is passed in which allows this eviction to be considered
+	response := testCtx.controller.MarkPodAsDeleted(testCtx.ctx, testNamespace, zoneAPod0.Name, "eviction-controller-test", NewMaxUnavailableZeroOverride(1))
+	require.NoError(t, response)
```
aknuds1: Naming the error `response` instead of the idiomatic `err` threw me off, and probably future readers:
```suggestion
	err := testCtx.controller.MarkPodAsDeleted(testCtx.ctx, testNamespace, zoneAPod0.Name, "eviction-controller-test", NewMaxUnavailableZeroOverride(1))
	require.NoError(t, err)
```

pkg/zpdb/eviction_controller_test.go
```diff
@@ -378,6 +415,55 @@ func TestPodEviction_MultiZoneClassic(t *testing.T) {
 	testCtx.controller.Stop()
 }
 
+// TestPodEviction_MultiZoneClassicOverrideHasNoEffectWhenNotZero validates that the MaxUnavailableZeroOverride
+// is only used when maxUnavailable=0. Setting the override when this is not true has no effect.
+func TestPodEviction_MultiZoneClassicOverrideHasNoEffectWhenNotZero(t *testing.T) {
+
```
aknuds1: [Nit] We don't start functions with a blank line:
```suggestion
```

PR #364, opened 2026-01-22
Support appending PR numbers in update-changelog-deps tool

andyasp:
This is an addition to a quality of life script used to help generate entries for `CHANGELOG.md` like:
```
* [ENHANCEMENT] Updated dependencies, including: #358 #363
  * `github.com/grafana/dskit` from `v0.0.0-20251208160410-c006bba17343` to `v0.0.0-20260119063514-a2b633bfa4ea`
  * `github.com/prometheus/common` from `v0.67.4` to `v0.67.5`
  * `github.com/sirupsen/logrus` from `v1.9.3` to `v1.9.4`
```

Previously this script didn't preserve existing PR numbers or add a new one, so the first line it generated would always be:
``` 
* [ENHANCEMENT] Updated dependencies, including:
```

This adds taking a PR number as a required argument for the Makefile command. If there are changes then the PR number is appended. The end result is I can now run it after renovate changes like https://github.com/grafana/rollout-operator/pull/358 and not have to manually fix the PR numbers after.

### DIFF
--- a/tools/update-changelog-deps/main.go
+++ b/tools/update-changelog-deps/main.go
@@ -69,6 +69,7 @@ func main() {
 	}
 
 	if len(changes) == 0 {
+		fmt.Fprintf(os.Stderr, "No changes were necessary\n")
 		return
 	}
 


andyasp:
The PR number won't be in the commit message when I want to run this (as a part of the PR itself) so I'll rework this to have the PR number be supplied instead.

PR #370, opened 2026-02-04
Deny evictions when cross-zone pod peers are not found

tcp13equals2:
This PR adds an additional test for pod eviction in partition aware environments.

A window of time exists where during an ingester scale-down, pods in zone-a can be shutdown and deleted and the zone-b StatefulSet not yet updated / scaling down.

During this time a voluntary eviction for a zone-b pod may be requested. 

Previously, if the equivalent zone-a pod had been shutdown and deleted then this eviction was allowed. As the StatefulSet was showing a heathy replica count and all the `found` pods for that partition where healthy.

This change requires that when there are no unhealthy or not ready pods in a peer partition we also test that the peer pod was actually found and considered. If the peer pod is not considered then the eviction will be denied.

Note - during a downscale the pods in the peer downscaled partitions should not be in the query/ingest path. So this an additional safeguard which should not be needed unless there is another issue in-play.

Note also that this will affect the `updateStatefulSetPods()` loop in the controller. In the case of a scale-up, zone-a will have pods which do not yet exist in zone-b. A zone-a pod eviction will be denied (including a delete for a version update) if the corresponding zone-b pod has not yet been provisioned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## main / unreleased
 
-[ BUGFIX ] Zone-aware PDB: deny evictions when cross-zone pod peers are not found in partitioned environments. #370
+* [BUGFIX] Zone-aware PDB: deny evictions when cross-zone pod peers are not found in partitioned environments. #370
+* [FEATURE] Add `grafana.com/rollout-force-replicas` annotation to manually override the replica count for a StatefulSet, bypassing the operator's normal scaling logic. #371
 
 ## v0.34.0
 
--- a/README.md
+++ b/README.md
@@ -112,6 +112,29 @@ Rollout-operator does NOT remember any state of "delayed scaledown" preparation.
 
 How is this different from `grafana.com/prepare-downscale` label used by `/admission/prepare-downscale` webhook? That webhook calls the "prepare-downscale" endpoint called *just* before the downscale is done, and pods are shutdown right after. On the other hand delayed downscale can take many hours. Delayed downscale and "prepare downscale" features can be used together.
 
+## Manual replica override
+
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
   * `k8s.io/apimachinery` from `v0.34.2` to `v0.35.0`
   * `k8s.io/client-go` from `v0.34.2` to `v0.35.0`
   * `sigs.k8s.io/controller-runtime` from `v0.22.4` to `v0.23.1`
+[ BUGFIX ] Zone-aware PDB: deny evictions when cross-zone pod peers are not found in partitioned environments. #370
```
andyasp: This release has already been cut so this is in the wrong section.

tcp13equals2: Ha ha - yes, thank you. Well spotted!

fionaliao (approved):
LTGM

pkg/zpdb/validator_partition_aware.go
```diff
@@ -47,10 +47,19 @@ func (v *validatorPartitionAware) considerSts(otherSts *appsv1.StatefulSet) bool
...
 
-func (v *validatorPartitionAware) accumulateResult(_ *appsv1.StatefulSet, r *zoneStatusResult) error {
+func (v *validatorPartitionAware) accumulateResult(sts *appsv1.StatefulSet, r *zoneStatusResult) error {
 	v.result.tested += r.tested
 	v.result.notReady += r.notReady
 	v.result.unknown += r.unknown
+
+	// If we were unable to confirm a pod status in this zone we should assume the result is unknown
+	// Note that this assumes that we only expect 1 pod per zone per partition.
+	// If ever this assumption changed, we would need to increment unknown by the number of expected pods per zone per partition.
```
fionaliao, line 57: For the 1 pod per zone per partition assumption, could you add a line to the README documenting it? 
It's probably unlikely to be change, but probably worth highlighting just in case.

PR #376, opened 2026-02-12
Update Go to 1.26 for v0.28 branch

andyasp:
Update Go to `1.26` for `v0.28`. Similar to https://github.com/grafana/rollout-operator/pull/375, except in the `v0.28` branch that is still being maintained for security releases.

PR #379, opened 2026-02-16
Replace k3d with kind for integration tests

andyasp:
Integration testing for `rollout-operator` was done using a wrapper package in `integration/k3t` which pulled in heavyweight dependencies to create a cluster per-test. This change eliminates those dependencies by instead relying on [kind](https://kind.sigs.k8s.io/) as a binary. The motivation is to reduce the maintenance burden these dependencies can cause. See the integration test failures in https://github.com/grafana/rollout-operator/pull/376 and https://github.com/grafana/rollout-operator/pull/377 for examples of this.

The basic idea is to: 
1. `kind create cluster --name <cluster name> --kubeconfig <temp file path>` at the start of a test, which runs a Docker container.
2. `kind load docker-image <image> --name <cluster name>` for `rollout-operator:latest` and `mock-service:latest` to bring in the images for the tests.
3. `kind delete cluster <cluster name>` at the end to clean up.

For transparency, I used AI to help with the initial lifting here. I also reviewed and rewrote code.

The integration tests pass locally for me with these changes.

Some notes on the changes:

- `kind` does not have an Ingress baked in like the `k3d` nodes did, but tests need to communicate with the mock services in the tests. There is a `cloud-provider-kind` that does have an Ingress, but instead of using that I worked around it by using port mappings for each `mock-service-{a,b,c}`.
- I noticed when testing these changes locally that (even before these changes) docker containers would get leaked if you hit Ctrl+C to interrupt a test. I added a signal handler in [test.Main](https://pkg.go.dev/testing#hdr-Main) to attempt cleanup on interrupt. It isn't perfect because it can race an in-progress `kind create cluster` call, but in practice it eliminated this problem for me.
- `getAndUpdateStatefulSetScale` was failing with a revision exception when run in `kind`. A revision exception was not the point of the test, which only wanted to update a value to see if that update got blocked, so I changed this to a `patch` instead.
- In `makeMockReady` I added waiting for an endpoint update. I think effectively this may have just added a bit more sleep time, but tests were consistently failing without it.
- There is a hardcoded `kindImage string = "kindest/node:v1.35.0"` that ideally gets maintained as we update kubernetes versions. We have a similar problem with tanka that got fixed in https://github.com/grafana/rollout-operator/pull/372. I think there's a way to get renovate to help with this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ build-image: clean ## Build the rollout-operator image
 .PHONY: build-image-boringcrypto
 build-image-boringcrypto: clean ## Build the rollout-operator image with boringcrypto
 	# Tags with the regular image repo for integration testing
-	docker buildx build --load --platform linux/amd64 --build-arg revision=$(GIT_REVISION) -t rollout-operator:latest -t rollout-operator:$(IMAGE_TAG) -f Dockerfile.boringcrypto .
+	docker buildx build --load --platform linux/$(GOARCH) --build-arg revision=$(GIT_REVISION) -t rollout-operator:latest -t rollout-operator:$(IMAGE_TAG) -f Dockerfile.boringcrypto .
 
 .PHONY: build-test-images
 build-test-images: build-image integration/mock-service/.uptodate ## Build both rollout-operator and mock-service images for testing


README.md
```diff
@@ -548,4 +548,22 @@ Note - `maxUnavailable` can be set to 0. In this case no voluntary evictions in
...
+
+## Development
+
+### Running Integration Tests
+
+Integration tests require Docker and kind (Kubernetes in Docker).
+
+To run all integration tests:
+```bash
+make build-image
```
tcp13equals2, line 561: Does a developer also need to install `kind`? 

Does a check for this in their path need to be added to the Makefile?

andyasp: > Does a developer also need to install kind?

Yes, you need it installed locally if you want to run the integration tests locally.

> Does a check for this in their path need to be added to the Makefile?

Failing faster on a check like that would be a nicer experience. I added a pre-check for `make integration`.

PR #422, opened 2026-04-20
rollout-operator forced delay for partition aware zpdb

tcp13equals2:
This PR introduces a new zpdb partition aware configuration option.

An optional `crossZoneEvictionDelay` (duration) has been added which when set will force this time to elapse following a pod being restarted (evicted) before another pod in the same partition can be evicted.

The time is calculated from when the pod returns to a ready + running state.

ie if `ingester-zone-a-0` is evicted, `ingester-zone-b-0` can not be voluntarily evicted until `ingester-zone-a-0` has been ready and running for at least the configured delay.

This delay is being added for ingesters so that we wait > 20 minutes between ingesters of different zones in order to get the correct active series numbers on a rollout.

To track the time that a pod returned to ready and running, an annotation is recorded on the pod. This allows the rollout-operator to be restarted without losing this value. As such, an additional permission has been added to the rollout-operator to allow it to patch pods.

Note that the existing recent pod eviction cache has not been updated to use a similar annotation - as the recent evictions cache only needs to hold records for a very short period of time prior to the pod eviction taking effect. 

---

> [!NOTE]
> **Medium Risk**
> Changes pod-eviction admission logic by adding time-based readiness gating and patching live Pod annotations, plus expands RBAC to allow `pods/patch`; mistakes could block legitimate evictions or increase API load.
> 
> **Overview**
> Adds a new `ZoneAwarePodDisruptionBudget.spec.crossZoneEvictionDelay` (Go duration string) that, in *partition-aware mode*, denies evictions until enough time has elapsed since the last pod in that partition returned to ready+running.
> 
> Implements this by maintaining a persistent `grafana.com/ready-time` Pod annotation via a new pod readiness tracker (with configurable patch timeout), wiring the readiness check into the partition-aware validator, and updating RBAC/manifests to allow `patch` on `pods`.
> 
> Updates CRDs/libsonnet, docs/runbooks, and adds unit + integration coverage (including a new e2e scenario) for delay enforcement and admission validation (reject delay without `podNamePartitionRegex`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd027c875b0fe95846b3a26c5922e7553f7cc366. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/zpdb/pod_eviction_cache.go
```diff
@@ -36,46 +38,69 @@ func newPodEvictionCache() *podEvictionCache {
...
 	c.entries[pod.Name] = podEvictionCacheValue{
 		expires:    expiresAt,
 		generation: pod.Generation,
+		force:      forced,
 	}
 }
 
 // hasPendingEviction returns true if this pod is in the cache and not expired
 func (c *podEvictionCache) hasPendingEviction(pod *corev1.Pod) bool {
 	// note that we do not clean up expired entries, as the assumption is the entry will be deleted shortly after being stored here
```
karlskewes, line 66: Is this still accurate? 
It looks like expired entries are now deleted per: https://github.com/grafana/rollout-operator/pull/422/files#diff-c60a327c221685b6f07c480c0882fdf85a48a06d2e8137efe4a92b276d8acf6eR73-R77 but _only_ if they are queried for again AFAICT.

pkg/zpdb/pod_eviction_cache.go
```diff
@@ -19,12 +19,14 @@ type podEvictionCacheValue struct {
 	expires time.Time
 	// the generation of the pod when added
 	generation int64
+	// force the record to stay in the cache until expiry
+	force bool
```
karlskewes, line 23: `force` is fine with the comment. 
An alternative might be `sticky` as in `sticky sessions` or similar.

pkg/zpdb/pod_eviction_cache_test.go
```diff
@@ -14,8 +15,28 @@ func TestLifecycle(t *testing.T) {
...
+	cache.recordEviction(pod, 5*time.Second)
+	require.True(t, cache.hasPendingEviction(pod))
+	// deleting the pod will not remove it from the cache until the force time expires
+	cache.delete(pod)
+	require.True(t, cache.hasPendingEviction(pod))
+
+	require.Eventually(t, func() bool {
+		cache.delete(pod)
+		return !cache.hasPendingEviction(pod)
+	}, 10*time.Second, 50*time.Millisecond)
```
karlskewes, line 41: `require.Eventually()` is used elsewhere in this codebase too. 
Could at some point use https://go.dev/blog/synctest#testing-time for more deterministic testing of timing.

karlskewes (dismissed):
LGTM, only a few observations not requiring action.

pkg/zpdb/pod_eviction_cache.go
```diff
@@ -70,7 +70,7 @@ func (c *podEvictionCache) hasPendingEvictionWithGeneration(pod *corev1.Pod) (bo
 	if !exists {
 		return false, 0
 	}
-	return exists, rec.generation
+	return exists && time.Now().Before(rec.expires), rec.generation
```
tcp13equals2, line 71: Note to reviewers - this time check was not previously here and it make sense for it to be here. 

This ensures that the cached record for the pod expires after a reasonable period of time. We only need this cached entry to persist long enough for the pod to be observed as shutting down.

This now matches the same cache test logic as `hasPendingEviction()`

pkg/zpdb/eviction_controller.go
```diff
@@ -292,10 +292,10 @@ func (c *EvictionController) HandlePodEvictionRequest(ctx context.Context, ar v1
 
 	if len(partition) > 0 {
 		// partition mode - the pod tallies are applied to all pods in other zones which relate to this partition
-		pdb = newValidatorPartitionAware(sts, partition, len(allStatefulSets.Items), pdbConfig, request.log)
+		pdb = newValidatorPartitionAware(sts, partition, len(allStatefulSets.Items), pdbConfig, c.podObserver.podEvictCache, c.podObserver.podReadinessCache, request.log)
 	} else {
 		// zone mode - we allow the eviction if no other zone is disrupted and the max unavailable within the eviction zone is not exceeded
-		pdb = newValidatorZoneAware(sts, len(allStatefulSets.Items))
+		pdb = newValidatorZoneAware(sts, len(allStatefulSets.Items), c.podObserver.podEvictCache)
```
tcp13equals2, line 299: Note for reviewers - although was have not touched the logic for the zone aware validator, the test which uses the eviction cache has been moved within the validator.

pkg/zpdb/pod_observer.go
```diff
@@ -112,15 +108,45 @@ func (c *podObserver) invalidatePodEvictionCache(obj interface{}, action string)
 }
 
 func (c *podObserver) onPodAdded(obj interface{}) {
-	c.invalidatePodEvictionCache(obj, "added")
+	pod, isPod := obj.(*corev1.Pod)
```
tcp13equals2, line 115: Note to reviewers - previously this casting check was in `invalidatePodEvictionCache()` - however now we different code paths from the observer interfaces to the  podReadinessCache these casting checks have been pulled into each observer callback function.

integration/e2e_test.go
```diff
@@ -476,6 +479,88 @@ func TestZoneAwarePodDisruptionBudgetPartitionMode(t *testing.T) {
 
 }
 
+// TestZoneAwarePodDisruptionBudgetPartitionModeWithCrossZoneEvictionDelay validates that a cross-zone eviction for the same partition
+// is denied while the crossZoneEvictionDelay has not expired, and then allowed after the delay has elapsed and the evicted pod has recovered.
+func TestZoneAwarePodDisruptionBudgetPartitionModeWithCrossZoneEvictionDelay(t *testing.T) {
```
charleskorn, line 484: It'd be good to test here that the other partition is not affected by the delay (eg. after `a-0` has restarted, but before the delay has elapsed, confirm that `a-1` or `b-1` can be evicted)

operations/rollout-operator/zone-aware-pod-disruption-budget.libsonnet
```diff
@@ -5,7 +5,8 @@
   // maxUnavailable - the max unavailable pod count. Can be expressed as an integer >= 0 or as a percentage 50% (0-100%)
   // podNamePartitionRegex - use to enable partition awareness. A regex which can extract a partition name from a pod name. ie [a-z\\-]+-zone-[a-z]-([0-9]+)
   // podNameRegexGroup - use to define the regex grouping which holds the partition name. default is 1
-  newZPDB(name, rolloutGroup, maxUnavailable, podNamePartitionRegex='', podNameRegexGroup=1):: {
+  // crossZoneEvictionDelay - optional Go duration (e.g. "20m") that must elapse before a pod in another zone for the same partition can be evicted. Requires podNamePartitionRegex.
+  newZPDB(name, rolloutGroup, maxUnavailable, podNamePartitionRegex='', podNameRegexGroup=1, crossZoneEvictionDelay=''):: {
```
charleskorn, line 9: Should this return an error if `crossZoneEvictionDelay` is set but `podNamePartitionRegex` isn't?

pkg/zpdb/pod_ready_cache.go
```diff
@@ -0,0 +1,133 @@
...
+	evicted bool
+	// the creationTimestamp of the pod - so we can detect stale pod updates
+	creationTimestamp int64
+}
+
+// podReadinessCache is a cache specific to tracking the time since
+// a pod returned to a ready + running state following an eviction.
+// Note that on process start-up, the time since the pod became ready/running
+// is not known. Only once the pod has been evicted at least once and returned
+// to service will this value be known.
```
charleskorn: Could this be read from the Kubernetes API on startup? Otherwise, the following series of events could cause the `crossZoneEvictionDelay` value not to be respected:

1. pod `a-1` is evicted
2. pod `a-1` restarts and becomes ready
3. rollout-operator restarts
4. eviction request for `b-1` arrives before delay has elapsed

If the time the pod became ready is not available, then what if we assumed that the pod became ready at the time the process started? Then at least we'll never evict another pod too early - it'll just wait longer to be allowed to be evicted.

This would help simplify the logic in `validatorPartitionAware.isReady` as well.

pkg/zpdb/pod_ready_cache.go
```diff
@@ -0,0 +1,133 @@
...
+// The consumer of the podReadinessCacheValue record for a pod can use the
+// `evicted` attribute to determine if the `since` time relates to when we first
+// observed the pod on startup vs the time since the pod returned
+// to service after an eviction.
+type podReadinessCache struct {
+	// pod name --> [ since time, creation timestamp ]
+	entries map[string]podReadinessCacheValue
```
charleskorn: When will pods be removed from this map? As far as I can see, this map will only ever grow, which means that over time it will accumulate all pods that have ever run since the rollout-operator process started.

pkg/zpdb/pod_ready_cache.go
```diff
@@ -0,0 +1,133 @@
...
+
+type podReadinessCacheValue struct {
+	// the time that this pod changed state
+	since time.Time
+	// the last observed state of the pod
+	readyRunning bool
+	// true when this pod has been evicted since we started observing it
+	evicted bool
+	// the creationTimestamp of the pod - so we can detect stale pod updates
+	creationTimestamp int64
```
charleskorn: Is there a reason not to use a `time.Time` here? If so, would be good to add a comment explaining why

pkg/zpdb/pod_ready_cache.go
tcp13equals2: Yes - but it will be keyed by the pod name which should be reasonably finite. 

If we are concerned about the size of this map we could apply the zpdb regex to the observed pods and only cache the pods which will be used by the zpdb.

charleskorn: Will this map as it currently stand capture pods from Deployments? If so, if there's a lot of scaling activity for those Deployments, then this will grow quite significantly given each new pod gets a unique name (unlike StatefulSets).

>we could apply the zpdb regex to the observed pods and only cache the pods which will be used by the zpdb

Can we use the labels from the ZPDB?

pkg/zpdb/pod_ready_cache.go
tcp13equals2: We could use the pods creation time - it wont be the same as being ready but may give some mitigation.

pkg/zpdb/pod_ready_cache.go
tcp13equals2: I can change to use time - no specific reason

pkg/zpdb/pod_ready_cache.go
charleskorn: I think we should err on the side of conservatism and use the current wall clock time as the time it became ready - otherwise it could evict a pod too early.

operations/rollout-operator/zone-aware-pod-disruption-budget.libsonnet
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/b3410b1f72cd359dbebc0ae48b9c3b31ac13d1ad

pkg/zpdb/pod_ready_cache.go
tcp13equals2: If we use the current wall clock time as the ready+running time then following a rollout-operator restart no voluntary evictions would be allowed until the delay expired.

Which is the "lesser evil" here - not allowing evictions for a period following a rollout-operator restart or allowing an eviction before the delay as expired. 

Given that the scenarios driving this functionality are asking for this as a nice to have and not because it's incident inducing I would err towards not blocking evictions after a rollout-operator restart.

Do you have a strong opinion on this topic?

pkg/zpdb/pod_ready_cache.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/5c220a661538edc1aea50b5a97148a32435d1226

pkg/zpdb/pod_ready_cache.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/63b6c30f273d257ac7059bb0b94ee3fa07270baa

pkg/zpdb/pod_ready_cache.go
tcp13equals2: Conversation moved offline ...

integration/e2e_test.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/3cff58c0e43d18878fead96ebfbfd4881854d735

pkg/zpdb/pod_ready_cache.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/40dad5f24cdb40234a015b2cc461abf000880262

docs/runbooks.md
```diff
@@ -153,3 +153,32 @@ ingester_rollout_pdb+:
...
+
+For instance;
+
+- t0 - `pod-zone-a-0` is evicted
+- t1 - `pod-zone-a-0` returns to full service (ready and running)
+- t2 - `pod-zone-b-0` is tested for eviction
+
+Assuming the PDB `max-unavailable` is 1, the `pod-zone-b-0` eviction will not be allowed until at least `crossZoneEvictionDelay` has elapsed since t1 (i.e. `t2 - t1 >= crossZoneEvictionDelay`).
+
+Note that this duration is calculated from when the `pod-zone-a-0` returned to service, not when it was evicted.
```
charleskorn: [nit] This might be less ambiguous:

```suggestion
Note that this duration is calculated from when the `pod-zone-a-0` becomes ready, not when it was evicted.
```

docs/runbooks.md
```diff
@@ -153,3 +153,32 @@ ingester_rollout_pdb+:
...
+
+When `crossZoneEvictionDelay` is unset or `0`, no delay is enforced and evictions follow the standard ZPDB logic.
+
+The time that a pod returned to service is recorded by setting a `grafana.com/ready` annotation on the pod. This ensures that
```
charleskorn: Same as above:
```suggestion
The time that a pod became ready is recorded by setting a `grafana.com/ready` annotation on the pod. This ensures that
```

docs/runbooks.md
```diff
@@ -153,3 +153,32 @@ ingester_rollout_pdb+:
...
+if the `rollout-operator` restarts the last ready time is not lost. The annotation is automatically removed when a pod transitions
+out of a ready+running state, so the delay is correctly re-measured from the next ready transition. The annotation is also lost if a pod is re-created.
+
+When the annotation is missing - e.g. the first time this version of the `rollout-operator` runs, or after a pod is re-created during
+a `rollout-operator` restart - the `rollout-operator` records `time.Now()` as the ready time on its first observation of the pod. The
```
charleskorn: [nit] `time.Now()` will only make sense to Go devs

```suggestion
a `rollout-operator` restart - the `rollout-operator` records the current time as the ready time on its first observation of the pod. The
```

pkg/admission/zpdb_validating_webhook_test.go
```diff
@@ -17,6 +17,12 @@ func TestZoneAwarePdbValidatorHandlerSuccess(t *testing.T) {
 	assertAllowResponse(t, request)
 }
 
+// TestZoneAwarePdbValidatorHandlerSuccessWithCrossZoneEvictionDelay tests with a valid configuration including crossZoneEvictionDelay
```
charleskorn, line 20: What about a ZPDB with `crossZoneEvictionDelay` set but not `podNamePartitionRegex`?

pkg/zpdb/pod_eviction_cache.go
charleskorn, line 58: Could this method be rewritten in terms of `hasPendingEvictionWithGeneration`?

pkg/zpdb/pod_observer.go
```diff
@@ -111,16 +109,59 @@ func (c *podObserver) invalidatePodEvictionCache(obj interface{}, action string)
 	c.podEvictCache.delete(pod)
 }
 
+// accept will return a pod and true of the given object is a pod and the is within the scope of our pdb
```
charleskorn:
```suggestion
// accept will return a pod and true if the given object is a pod and the is within the scope of our pdb
```

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+
+const (
+	// podReadyAnnotationKey is set on a pod to record when it was first observed as ready and running.
+	// The value is the observation time formatted as RFC3339 (UTC).
+	// Once set, the annotation is preserved across subsequent ready observations so the original
+	// timestamp is retained. The annotation is removed when the pod transitions out of a ready+running state.
+	podReadyAnnotationKey = "grafana.com/ready"
+
+	// podReadyAnnotationPatchTimeout bounds the kube API call that maintains podReadyAnnotationKey.
+	podReadyAnnotationPatchTimeout = 5 * time.Second
```
charleskorn: Should this be a CLI flag?

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+	lock.Lock()
+	defer lock.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), podReadyAnnotationPatchTimeout)
+	defer cancel()
+
+	var patch string
+
+	if util.IsPodRunningAndReady(pod) {
+		// Get the latest version of this pod - avoid operating on stale observations
```
charleskorn: This still won't completely mitigate operating on stale information - is there a way to make the `Patch` call below conditional, so we only update the pod if it hasn't changed from what was passed as `pod` here?

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+		if value, ok := current.Annotations[podReadyAnnotationKey]; ok && value != "" {
+			return
+		}
+
+		// set the ready annotation value to Now()
+		patch = fmt.Sprintf(`{"metadata":{"annotations":{%q:%q}}}`, podReadyAnnotationKey, time.Now().UTC().Format(time.RFC3339))
+
+	} else {
+		// null out the ready annotation
+		patch = fmt.Sprintf(`{"metadata":{"annotations":{%q:null}}}`, podReadyAnnotationKey)
```
charleskorn: This feels quite brittle - any reason not to marshal a Go struct as JSON here?

pkg/zpdb/validator_partition_aware.go
```diff
@@ -77,3 +84,29 @@ func (v *validatorPartitionAware) successMessage() string {
...
+	// transitioned to ready+running. The transition time is read from the pod's
+	// podReadyAnnotationKey annotation (or time.Now() when the annotation is absent or invalid,
+	// which is the safe default that denies eviction until the delay window has been observed).
+	now := time.Now()
+	since := v.readyTracker.get(pod)
+	if now.After(since.Add(v.pdbConfig.crossZoneEvictionDelay)) {
+		return true
+	}
+
+	level.Info(v.log).Log("msg", "Pod not considered ready - not enough time has elapsed since this pod became ready", "pod", pod.Name, "time-until-ready", since.Add(v.pdbConfig.crossZoneEvictionDelay).Sub(now))
```
charleskorn: If `v.readyTracker.get()` returns `time.Now()` because the pod hasn't been seen, it might be useful to log a different message here to make it clearer that's what's happened

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+// restarts. The annotation on the pod itself is the source of truth; this struct only
+// reflects observed pod state into that annotation.
+type podReadinessTracker struct {
+	logger     log.Logger
+	kubeClient kubernetes.Interface
+	namespace  string
+
+	// locks by pod name
+	podLocks   map[string]*sync.Mutex
+	podLocksMu sync.Mutex
```
charleskorn: Why do we need these locks?

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/grafana/rollout-operator/pkg/util"
+)
+
+const (
+	// podReadyAnnotationKey is set on a pod to record when it was first observed as ready and running.
+	// The value is the observation time formatted as RFC3339 (UTC).
+	// Once set, the annotation is preserved across subsequent ready observations so the original
+	// timestamp is retained. The annotation is removed when the pod transitions out of a ready+running state.
+	podReadyAnnotationKey = "grafana.com/ready"
```
charleskorn: [nit] Perhaps this would be a clearer name?
```suggestion
	podReadyAnnotationKey = "grafana.com/ready-time"
```

karlskewes: Perhaps: `podReadyAnnotationKey = "grafana.com/ready-timestamp"` to avoid confusion between duration and timestamp.

docs/runbooks.md
```diff
@@ -153,3 +153,32 @@ ingester_rollout_pdb+:
...
+- t1 - `pod-zone-a-0` returns to full service (ready and running)
+- t2 - `pod-zone-b-0` is tested for eviction
+
+Assuming the PDB `max-unavailable` is 1, the `pod-zone-b-0` eviction will not be allowed until at least `crossZoneEvictionDelay` has elapsed since t1 (i.e. `t2 - t1 >= crossZoneEvictionDelay`).
+
+Note that this duration is calculated from when the `pod-zone-a-0` becomes ready, not when it was evicted.
+
+When `crossZoneEvictionDelay` is unset or `0`, no delay is enforced and evictions follow the standard ZPDB logic.
+
+The time that a pod became ready is recorded by setting a `grafana.com/ready` annotation on the pod. This ensures that
```
charleskorn:
```suggestion
The time that a pod became ready is recorded by setting a `grafana.com/ready-time` annotation on the pod. This ensures that
```

charleskorn:
cursor review

pkg/admission/zpdb_validating_webhook_test.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/e85719ca7846e67ff8c9b74b32f8ced02b0d2778

pkg/zpdb/pod_eviction_cache.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/c5d30fa6cb3d24003135169bc9f1290f5b08a81c

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/d0678b76458a3de30dbd3789987a68eb114cefa7

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: I don't think we can use the conditional update functionality in the patch - from what I read that will only work to set an existing field. It would not work if the field was not already set. 

But what is risk and impact of operating on stale information?

* this pod_readiness_tracker is the only thing reading and writing these annotations - and it's running within the rollout_operator which is a singleton / single pod
* the code path is serialized per pod and is reached via the informer handler which runs in a single threaded routine. 

So the chance of the annotation being concurrently updated should never happen.

Further more, if a stale write was done (state changed after the informer update or between testing ready/running and the patch call) - the annotation write triggers the pod informer and we loop back to call this `observed()` function again - where we re-test the pod readiness and update the annotation again if it's wrong.  

Am I missing something else here?

I have refined this function and comments - https://github.com/grafana/rollout-operator/pull/422/changes/2720134034e13ca4d47a67f4c7747ecef6b90c88

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/a54802df1df20e58dcf494670ea8a22ac938a08b

pkg/zpdb/validator_partition_aware.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/b1a2b88618800d90b41745b580c62e8447d38876

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/9654b1003f74c2c323f23e6287646d893821006e

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,133 @@
...
+
+	} else {
+		// null out the ready annotation
+		patch = fmt.Sprintf(`{"metadata":{"annotations":{%q:null}}}`, podReadyAnnotationKey)
+	}
+
+	if _, err := c.kubeClient.CoreV1().Pods(c.namespace).Patch(ctx, pod.Name, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{}); err != nil {
+		level.Warn(c.logger).Log("msg", "failed to patch pod ready annotation", "pod", pod.Name, "err", err)
+	}
+}
```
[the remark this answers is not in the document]
tcp13equals2: There are now 2 fast-paths for avoiding the patch if not required.

tcp13equals2:
@cursor push bc06e9b7a6

operations/rollout-operator/rollout-operator.libsonnet
```diff
@@ -75,6 +75,7 @@
     'kubernetes.namespace': $._config.namespace,
     'use-zone-tracker': true,
     'zone-tracker.config-map-name': 'rollout-operator-zone-tracker',
+    'zpdb.pod-ready-annotation-patch-timeout': '5s',
```
charleskorn, line 78: [nit] Do we need to set this given it's the default value?

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,139 @@
...
+	// value as a delete). A non-nil value means "set the annotation to *value".
+	var value *string
+
+	if util.IsPodRunningAndReady(pod) {
+		// Fast path: the informer's in-memory pod copy already shows the annotation, so leave
+		// the live pod untouched.
+		// If there are further changes to the pod we would expect the informer to fire again
+		// and observed() called. Similarly, when the patch is applied this will trigger the
+		// informer and observed() to run again ensuring annotation convergence.
+		if v, ok := pod.Annotations[podReadyAnnotationKey]; ok && v != "" {
```
charleskorn: Is it valid for an annotation to have an empty value? We don't do the same check below

pkg/zpdb/pod_readiness_tracker.go
```diff
@@ -0,0 +1,139 @@
...
+				// annotation found and parsed: pod is ready+running, return the recorded since time
+				return true, true, since
+			}
+		}
+		// annotation missing, empty or unparseable: pod is ready+running, fall back to now
+		return false, true, time.Now()
+	}
+
+	// pod is not ready+running: nothing to report, fall back to now
+	return false, false, time.Now()
```
charleskorn: [nit] It might be clearer to reverse the first condition to reduce the nesting:

```go
if !util.IsPodRunningAndReady(pod) {
	// pod is not ready+running: nothing to report, fall back to now
	return false, false, time.Now()
}

... old if body
```

charleskorn (commented):
LGTM modulo question about empty annotations

operations/rollout-operator/rollout-operator.libsonnet
tcp13equals2: I specifically put it in to make it more obvious that we have this value available to be set. Operators can see this in the libsonnet or if they are looking at a kube manifest file.

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/ed18f328ff7473ea1c6c489729a59a80c83c0541

pkg/zpdb/pod_readiness_tracker.go
tcp13equals2: https://github.com/grafana/rollout-operator/pull/422/changes/d1a927d4888df884e4b3c14c70e8b8032b8d57e3

charleskorn (dismissed):

charleskorn (approved):

PR #425, opened 2026-04-27
Backport TestOTELTracingInitialization to 0.28.x

andyasp:
Backports the test originally added in https://github.com/grafana/rollout-operator/pull/408 to prevent the same issue from resurfacing in this branch when dependencies are updated.


alexweav (approved):

andyasp:
I'm going to double check this more closely because I think the deployment may not be getting the environment variables the tests set correctly.

andyasp:
There we go, let's see if updating dskit fixes it.

andyasp:
That worked.

PR #441, opened 2026-06-04
Fix Kubernetes API client rate limiter runaway under high load

pracucci:
## The issue

Under a burst of highly concurrent pod eviction requests (for example when Kubernetes wants to drain many nodes at once), the rollout-operator's admission webhooks started failing with errors like:

```
client rate limiter Wait returned an error: context canceled
```

The operator's Kubernetes API request throughput collapsed to near zero and stayed there for a sustained period, even though the actual rate of requests it was issuing was very low (well under the configured client QPS). Pod evictions were denied for the whole duration, and the `context canceled` error made it look like a server-side timeout rather than self-inflicted client-side throttling.

Example log during the issue:

```
ts=2026-06-02T11:45:27.373453875Z method=admission.PodEviction() object.name=store-gateway-zone-a-605 object.resource=pods object.namespace=cortex-dedicated-14 request.uid=e32630b1-7392-4b98-8f40-adab1cb22c04 request.dry_run=false level=error msg="pod eviction denied" reason="unable to find pod by name" err="client rate limiter Wait returned an error: context canceled"
```

Screenshot 2026-06-04 at 14 19 20

Per-second rate of ZPDB eviction requests:

Screenshot 2026-06-04 at 14 20 26

Per-second rate of all K8S API calls (here you can see that it drops to a 0.01/s, even if the per API group rate limiter is set to 5 new requests / s):

Screenshot 2026-06-04 at 14 20 57

## Root cause

The rollout-operator throttles its calls to the Kubernetes API with client-go's client-side rate limiter (a token bucket). Two things combined to make it stall:

1. **Webhook request contexts have no deadline.** Admission webhook handlers issue their Kubernetes API calls under the incoming HTTP request context. `net/http` cancels that context when the connection closes, but never sets a *deadline* on it. client-go calls `rateLimiter.Wait(ctx)` *before* applying its own per-request timeout, so the limiter always sees a deadline-less context.

2. **The token bucket runs away under concurrency.** With no deadline, the token bucket (`golang.org/x/time/rate`) never rejects a request: it reserves a token arbitrarily far in the future and parks the goroutine. When the apiserver gives up on the webhook (after its webhook timeout) and closes the connection, the request is cancelled. Cancelling a reservation only returns its tokens if no later reservation was made after it, and under concurrent load there is always a later reservation. So the cancelled requests restore nothing, the bucket's "reserved until" timeline races ahead of real time, and every subsequent request reserves even further into the future, gets cancelled, and restores nothing. The effective throughput collapses toward zero (far below the configured QPS) and every request fails at the limiter with `context canceled`.

A related, secondary problem: client-go's rate limiter is per-API-group only by accident. When QPS is left unset, each API group gets its own token bucket; but as soon as a custom QPS is set, client-go uses a single bucket shared across all API groups. So there was no way to raise the per-group limit without collapsing everything into one global limit.

### (a) Enforce a context deadline before the rate limiter — the actual fix

Admission webhook handlers now run under a context with a finite deadline (a fraction of the webhook server request timeout, `-server-tls.request-timeout`). This is what stops the runaway.

The key is *how* the token bucket behaves with a deadline. When `Wait(ctx)` is given a context with a deadline, the bucket computes how long the request would have to wait for a token and, if that exceeds the deadline, it **rejects the request immediately without mutating its internal state**: it does not reserve a token, so it does not push the reservation timeline forward. With a deadline-less context the same request would instead reserve a far-future token, park, get cancelled, and (under concurrency) restore nothing, poisoning the bucket for everyone. So the deadline turns "reserve, then stall, then collapse" into "fail fast and keep serving at the configured rate": over-budget requests are rejected quickly with a clear error while in-budget requests keep flowing at QPS. As a bonus, the deadline also makes in-flight API calls cancel promptly instead of running on past the point where the apiserver has already given up.

The test `TestServe_HandlerDeadlinePreventsRateLimiterStall` shows the behaviour difference.

### (b) A custom per-API-group rate limiter (and `PodHTTPClient`)

We replace client-go's built-in limiter with our own, implemented at the HTTP transport level, so we always get an independent token bucket *per API group*, even when a custom QPS/burst is configured (something client-go cannot do, since a custom QPS there becomes a single global bucket). client-go's own limiter is disabled and the throttling is done in the transport, keyed by the API group parsed from the request path. `-kubernetes.client-qps` / `-kubernetes.client-burst` configure the per-group limit.

Because the operator also makes direct HTTP calls to **pod** endpoints (the prepare-downscale endpoints), and those must never be throttled by the Kubernetes API rate limiter (they can legitimately run well above the API QPS), we introduced a dedicated `PodHTTPClient` type. Every code path that calls a pod endpoint now uses it. It is a distinct type (so the Kubernetes API client cannot be passed to it by mistake) backed by a transport completely separate from the Kubernetes API client, so pod-endpoint calls bypass the rate limiter entirely.

### (c) A dedicated Kubernetes client per webhook

Each admission webhook now gets its own Kubernetes client, with its own rate-limiter buckets. This isolates them: a single overloaded webhook can only exhaust its own buckets and can no longer throttle the other webhooks or the core controller's reconcile loop.


cmd/rollout-operator/main.go
```diff
@@ -191,28 +196,32 @@ func main() {
...
 		level.Warn(logger).Log("msg", msg, "method", method, "cluster_validation_label", cfg.kubeNamespace)
 		metrics.ClientInvalidClusterValidationLabelRequests.WithLabelValues(method, "http", cfg.kubeNamespace).Inc()
 	}
-	httpRT = middleware.ClusterValidationRoundTripper(cfg.kubeNamespace, reporter, httpRT)
+
+	// Client for the HTTP endpoints the operator calls directly on pods (e.g. prepare-downscale). It uses a
+	// transport separate from the Kubernetes API client, so these calls are never subject to the
+	// per-API-group rate limiter and are tracked under their own metric. The 5s timeout matches the value
+	// previously hardcoded in the prepare-downscale webhook.
+	podHTTPClient := instrumentation.NewPodHTTPClient(nil, cfg.kubeNamespace, reporter, 5*time.Second, reg)
```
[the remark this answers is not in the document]
pracucci: Intentional. These prepare-downscale endpoints are fast operations (they just record/return a timestamp), and the prepare-downscale webhook has already used a 5s timeout for the same endpoints in production. The previous 5m bound on the controller path was incidental — those calls happened to share the Kubernetes API client — not a deliberate choice.

`-kubernetes.client-timeout` governs calls to the Kubernetes API server (API and resource/subresource endpoints), not direct HTTP calls to endpoints exposed by pods. Those now go through a dedicated `PodHTTPClient` precisely to decouple them from the Kubernetes API client (separate transport, not subject to the per-API-group rate limiter, with its own timeout). If a pod call does exceed 5s, the error is logged and the downscale is retried on the next reconcile, so it isn't a silent failure.

cmd/rollout-operator/main.go
```diff
@@ -88,8 +89,8 @@ func (cfg *config) register(fs *flag.FlagSet) {
 	fs.StringVar(&cfg.kubeAPIURL, "kubernetes.api-url", "", "The Kubernetes server API URL. If not specified, it will be auto-detected when running within a Kubernetes cluster.")
 	fs.StringVar(&cfg.kubeConfigFile, "kubernetes.config-file", "", "The Kubernetes config file path. If not specified, it will be auto-detected when running within a Kubernetes cluster.")
 	fs.DurationVar(&cfg.kubeClientTimeout, "kubernetes.client-timeout", 5*time.Minute, "HTTP client timeout. This applies to requests issued to both the Kubernetes API and Kubernetes resource endpoints.")
```
tcp13equals2: [nit] - Is it worth updating this comment to make it clear that this does not apply to calls made directly on pods?

cmd/rollout-operator/main.go
```diff
@@ -99,6 +100,7 @@ func (cfg *config) register(fs *flag.FlagSet) {
 	fs.IntVar(&cfg.serverTLSPort, "server-tls.port", 8443, "Port to use for exposing TLS server for webhook connections (if enabled).")
 	fs.StringVar(&cfg.serverCertFile, "server-tls.cert-file", "", "Path to the TLS certificate file if not using the self-signed certificate.")
 	fs.StringVar(&cfg.serverKeyFile, "server-tls.key-file", "", "Path to the TLS private key file if not using the self-signed certificate.")
+	fs.DurationVar(&cfg.serverTLSRequestTimeout, "server-tls.request-timeout", 10*time.Second, "Time budget to handle a single TLS server (admission webhook) request: used as the server read and write timeouts, and (at 90%) as the deadline imposed on the request context passed to the webhook handler. Should be smaller than the apiserver's webhook timeout so the rollout-operator responds before the apiserver gives up.")
```
tcp13equals2, line 105: [nit] - "Should be **smaller** than the apiserver's webhook timeout" - is this 10 second default equal the default Kubernetes ValidatingWebhookConfiguration timeout? I think the default timeout for a ValidatingWebhookConfiguration is 10 seconds.

tcp13equals2 (dismissed):
Approved with a couple of minor nits for your consideration.

cmd/rollout-operator/main.go
pracucci: Good point. Made the pod-call timeout configurable via the new `-pods.client-timeout` flag (default 5s) and updated the help text to clarify it's separate from `-kubernetes.client-timeout`.

cmd/rollout-operator/main.go
pracucci: Yes, the apiserver's default webhook `timeoutSeconds` is 10s, same as this default. The margin actually comes from the 90% factor: the deadline imposed on the handler context is 90% of this value (9s by default), so admission logic is bounded to ~9s and the operator returns before the apiserver's 10s timeout. The "should be smaller than the apiserver's webhook timeout" wording is meant as "don't set it *larger*" — keeping the default at 10s since the 90% headroom covers it.

pkg/admission/serve.go
```diff
@@ -35,9 +36,20 @@ func delegateV1beta1AdmitToV1(f AdmitV1Func) admitv1beta1Func {
...
 	return func(w http.ResponseWriter, r *http.Request) {
+		// net/http request contexts have no deadline (they're only cancelled when the connection closes),
+		// but the client-side rate limiter needs one to bound its wait — otherwise concurrent webhook
+		// bursts make its token bucket stall.
+		ctx := r.Context()
+		if handlerTimeout > 0 {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, handlerTimeout)
+			defer cancel()
+		}
```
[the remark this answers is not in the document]
pracucci: Intentional. `-server-tls.request-timeout` is the budget to fully handle a request — body read included — so the operator responds before the apiserver's webhook timeout gives up. Starting the deadline only after the body read would let a slow read push total response time past the apiserver timeout, which is exactly what this budget is meant to bound. The body read is also already bounded by the TLS server's read timeout (set to the same value), and the 90% factor on the handler context leaves headroom to write the response.

tcp13equals2 (approved):
LGTM Marco. I'll approve but I am guessing you will need a re-approval once those linting errors are addressed.

pracucci:
> LGTM Marco. I'll approve but I am guessing you will need a re-approval once those linting errors are addressed.

CI is failing because of... https://www.githubstatus.com/incidents/m7n7sm0sr1pz

PR #464, opened 2026-07-21
Split integration tests across CI runners

julienduchesne:
## Summary

Integration jobs currently run the full suite on one runner (~16m). Shard them across multiple runners so CI wall time stays manageable as the suite grows.

- Matrix `integration` and `integration-boringcrypto` over two shards by default
- `make integration` honors `SHARD_INDEX` / `SHARD_COUNT` (local runs unchanged when unset)
- Runner count is the length of `matrix.shard` (`SHARD_COUNT` comes from `strategy.job-total`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,8 @@ SHARD_COUNT ?= 1
 .PHONY: integration
 integration: ## Run integration tests (set SHARD_INDEX/SHARD_COUNT to split across runners)
 integration: check-kind integration/mock-service/.uptodate
-	@set -euo pipefail; \
+	@# bash: make uses /bin/sh (dash on Ubuntu), which has no pipefail
+	@bash -euo pipefail -c '\
 	if [ "$(SHARD_COUNT)" -le 1 ]; then \
 		go test -v -tags requires_docker -count 1 -timeout 1h ./integration/...; \
 	else \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -87,10 +87,34 @@ test-boringcrypto: ## Run tests with GOEXPERIMENT=boringcrypto
...
+SHARD_INDEX ?= 0
+SHARD_COUNT ?= 1
+
 .PHONY: integration
-integration: ## Run integration tests
+integration: ## Run integration tests (set SHARD_INDEX/SHARD_COUNT to split across runners)
 integration: check-kind integration/mock-service/.uptodate
-	go test -v -tags requires_docker -count 1 -timeout 1h ./integration/...
+	@# bash: make uses /bin/sh (dash on Ubuntu), which has no pipefail
+	@bash -euo pipefail -c '\
```
andyasp: This feels like it would be cleaner if it was an actual bash file under `tools/` instead.

andyasp (commented):
Sharding makes sense to me. For context, a major contributor to the integration test time is that a `kind` cluster is created and torn down for each one individually. That isn't always necessary and I suspect that avoiding it when it's not would give a significant speed increase. That is a far more involved change though.

Makefile
julienduchesne: Done!

andyasp (approved):
Thanks!

PR #468, opened 2026-07-22
Fix prepare-downscale path joining for leading slashes

### LINKED ISSUE #287: grafana.com/prepare-downscale-http-path should either follow redirects or error on non 2xx response
When configuring this for Loki scale down i incorrectly set `/ingester/prepare_shutdown` as the prepare-downscale-http-path annotation value, which led to the rollout operator calling `<pod fqdn>:3100//ingester/prepare_shutdown`, which in turn returns a 301 to the correct uri. The rollout operator does not follow that 301, but then continues with the scale down. In the context of Loki, this means that the ingesters do not leave the ring cleanly and become unhealthy.

Ultimately the operator should fail if it does not receive a 2xx response to the call. Ideally it should follow 3xx responses until either it gets a 2xx or 4xx/5xx response and a conclusive success / fail state can be determined.

I am also of the opinion that the url should only have a `/` prepended when it is a relative path.

julienduchesne:
## Summary

Stop prepare-downscale from building broken pod URLs when `grafana.com/prepare-downscale-http-path` includes a leading `/`.

A leading slash previously produced URLs like `host:port//ingester/prepare_shutdown`. Many servers answer that with a 301, and Go's HTTP client then rewrites POST to GET — so the downscale could proceed without actually preparing the pod.

- Join the annotation path without doubling a leading `/`, and `path.Clean` duplicate slashes

Fixes: https://github.com/grafana/rollout-operator/issues/287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   * `k8s.io/apiextensions-apiserver` from `v0.36.2` to `v0.36.3`
   * `k8s.io/apimachinery` from `v0.36.2` to `v0.36.3`
   * `k8s.io/client-go` from `v0.36.2` to `v0.36.3`
+* [BUGFIX] Avoid doubling a leading `/` when building prepare-downscale pod URLs from `grafana.com/prepare-downscale-http-path`. #468
 
 ## v0.38.1
 
--- a/pkg/admission/prep_downscale.go
+++ b/pkg/admission/prep_downscale.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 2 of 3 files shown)


pkg/admission/prep_downscale.go
```diff
@@ -458,23 +460,37 @@ func checkReplicasChange(logger log.Logger, oldInfo, newInfo *objectInfo) *admis
...
+// collapses duplicates so values like "//ingester/prepare_shutdown" do not produce a double-slash URL
+// (which many servers answer with a 301).
+func prepareDownscalePath(p string) string {
+	if p == "" {
+		return "/"
+	}
+	if !strings.HasPrefix(p, "/") {
+		p = "/" + p
+	}
+	return path.Clean(p)
```
andyasp: This introduces a behavior change that would strip a "/" suffix. That seems relatively low risk to me though and I don't see an example of that documented or in our own use.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## main / unreleased
 
+* [BUGFIX] Avoid doubling a leading `/` when building prepare-downscale pod URLs from `grafana.com/prepare-downscale-http-path`. #287
```
andyasp:
```suggestion
* [BUGFIX] Avoid doubling a leading `/` when building prepare-downscale pod URLs from `grafana.com/prepare-downscale-http-path`. #468
```

andyasp (commented):
LGTM, but the changelog entry should reference the PR number and not the issue.

pkg/admission/prep_downscale.go
julienduchesne: You're right, and I think a trailing slash has some legit usecases. Removed path.Clean

andyasp (approved):
Thanks!

PR #473, opened 2026-07-28
Add read-only /ui/status page for StatefulSet rollouts

julienduchesne:
## Summary

Add a small embedded status page so operators can see rollout-group progress without scraping metrics or describing every StatefulSet by hand.

- Serve read-only HTML at `/ui/status` on the existing HTTP port (`-server.port`, default 8001)
- Build snapshots from informer caches: group/zone phase, ready replicas, revisions, and updated pods
- Local preview with demo data via `make status-ui-preview`
- Keep presentation separate from status collection so future health gates, phased Deployments, and authenticated actions can plug in later without shipping them now

image


koloss2001:
Could you clarify what `rollout-paused` should do when the paused zone is not Ready?

Assume the target version is `R2`, and we have two zones.

Example 1:

```text
zone A: R1, R1, R2 — paused — 2/3 Ready
zone B: R1, R1, R1 — Ready
```

The existing test expects the operator to skip A and start rolling B. This means we may disrupt B while A is already unhealthy.

Example 2:

```text
zone A: R2, R2, R2 — paused — 2/3 Ready
zone B: R1, R1, R1 — Ready
```

The controller still skips A and tries to roll B. However, the new status UI says B is waiting for A.

I expected pause to mean “do not delete pods in A.” I did not expect it to mean “ignore A’s readiness when deciding whether B can roll.”

Which behavior do we want?

- If pause is meant to ignore readiness and continue to B, the UI should match that, and this safety impact should be documented.
- If a not-Ready zone should always block the next zone, the controller behavior and the existing test should change.

Could we also add a test for the second example?

PR #475, opened 2026-07-30
Add CRD compatibility CI check

julienduchesne:
## Summary

Add a CRD compatibility check so PRs cannot accidentally introduce backward-incompatible schema changes.

- Compare both operator CRDs against the exact PR base commit using `crdify` v0.6.0
- Expose a local Make target with a configurable base ref

Made with [Cursor](https://cursor.com)


Makefile
```diff
@@ -29,6 +29,13 @@ DOC_SOURCES_PATH := docs
 
 REGO_POLICIES_PATH=operations/policies
 
+# CRDs checked for backward-incompatible schema changes.
+CRD_PATHS := \
```
tcp13equals2: Any reason not to pick up all yaml files in this directory? So if new CRDs are added they are picked up

julienduchesne: You're right. Done!

tcp13equals2 (approved):
LGTM

PR #479, opened 2026-08-10
Pod status from informer cache vs live API call

tcp13equals2:
In large environments the pod eviction code path can exhaust the rollout-operator's client-side rate limiter when looking up pod and StatefulSet status. This happens when there is a large eviction event, such as high node pressure causing mass evictions.

The hot path for each eviction is listing every pod in each zone of the rollout group and determining their status. For a two-zone group that is three `core/v1` calls per eviction — one Pod `Get` for the pod being evicted, plus one Pod `List` per zone — and two `apps/v1` calls.

Once the bucket is saturated, requests queue until they exceed the webhook deadline and fail, and those failures are retried, which sustains the saturation. In-flight eviction requests settle at roughly `QPS × webhook deadline` and stay there.

This PR does the following;
* small modification to the eviction handler to avoid an API call to lookup a pod if it's not within our PDB scope. We first lookup the pod to be evicted in the informer cache and if found test it to see if it falls within our scope. Since this determination is based off it's labels, it's safe to use the cached informer object. This will avoid an API call being made for every pod in a mass eviction and reduce the risk of rate limiting.
* added alert rule to notify when rate limiting is being applied
* updated runbook with steps to mitigate rate limiting / change rate limiting rules
* updated dashboard mixin to show rate limiting and show a desired QPS setting


operations/rollout-operator-mixin/dashboards/rollout-operator.libsonnet
```diff
@@ -152,6 +167,67 @@ local filename = 'rollout-operator.json';
...
+      .addPanel(
+        local title = 'Percent of requests rejected by rate limiter';
+
+        $.timeseriesPanel(title) +
+        $.panelDescription(title, 'Share of Kubernetes API requests per API group which never left the client because the client-side rate limiter had no token available. There is no healthy non-zero level for this.') +
+        $.queryPanel(
+          local selector = $.rolloutOperator_jobMatcher();
+          local rejected = 'sum by (api_group) (rate(rollout_operator_kubernetes_api_client_rate_limited_requests_total{%s}[$__rate_interval]))' % [selector];
+          local attempted = 'sum by (api_group) (rate(rollout_operator_kubernetes_api_client_request_duration_seconds_count{%s}[$__rate_interval]))' % [selector];
+          '%s / (%s + %s)' % [rejected, rejected, attempted],
```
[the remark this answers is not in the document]
tcp13equals2: https://github.com/grafana/rollout-operator/pull/479/commits/224ab1c2f189a931a7c27c9e7859377b5b5659c9

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main / unreleased
```
karlskewes, line 4: Can you please add an entry for the behaviour change to lookup in pod informer cache first.

tcp13equals2: https://github.com/grafana/rollout-operator/pull/479/commits/abdcf3850e10bf8b48651a20d59e42696022e725

operations/rollout-operator-mixin/alerts/alerts.libsonnet
```diff
@@ -41,6 +41,39 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+        {
+          // Compares each pod's own throughput against its own configured limit, since the token bucket is
+          // per-process. rollout_operator_kubernetes_api_client_rate_limit_qps is only published while rate
+          // limiting is enabled, so this alert is naturally absent - not a divide-by-zero - where it isn't.
+          alert: $.alertName('KubernetesAPIClientApproachingRateLimit'),
+          expr: |||
+            sum by (%(instance_labels)s, api_group) (rate(rollout_operator_kubernetes_api_client_request_duration_seconds_count[5m]))
+            / on (%(instance_labels)s) group_left()
+            max by (%(instance_labels)s) (rollout_operator_kubernetes_api_client_rate_limit_qps)
+            > 0.8
```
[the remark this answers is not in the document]
tcp13equals2: https://github.com/grafana/rollout-operator/pull/479/commits/c4639cc05855cd8d7eb5fe3571565ca5ec7c0f5f

operations/rollout-operator-mixin/alerts/alerts.libsonnet
```diff
@@ -41,6 +41,43 @@ local utils = import 'mixin-utils/utils.libsonnet';
             message: 'A sustained number of inflight ZPDB eviction requests has been observed.',
           },
         },
+        {
+          // There is no healthy non-zero level for this counter: a request which fails to acquire a token is
+          // never sent. Any sustained rate means the client-side rate limiter has become the bottleneck, so
+          // this alerts on > 0 held for long enough to rule out a momentary burst.
+          alert: $.alertName('KubernetesAPIClientRateLimited'),
+          expr: 'sum by (namespace, api_group) (rate(rollout_operator_kubernetes_api_client_rate_limited_requests_total[5m])) > 0',
```
karlskewes, line 49: I think it's right to leave `component` out of the labels here otherwise we could end up cycling through components without any single component firing for the full 15m. The fate is shared.
Can investigate the alert by querying full metrics.

karlskewes (approved):

